PR #7, opened 2014-06-22
Change configure scripts to allow out-of-tree builds.

stephankramer:
This allows you to build fluidity in a separate directory (including all
object files and libraries), without touching any of the checked in source
directories at all. This makes it possible to switch between branches in
a git clone of fluidity without loosing the various associated builds.

Simply call configure from any directory (other than the fluidity top
directory). The Makefiles will then be created in a hierarchy of directories
created under the current directory. When running make subsequently all new
files that are created during the build (object files, libraries, binaries,
etc.) are placed inside that new hierarchy.

For example to build fluidity in a separate build/ directory, inside the
fluidity checkout do:

   mkdir build; cd build
   ../configure && make

You can check using "git status" that indeed no new files (other than build/)
have been created. Your fluidity executable is build/bin/fluidity.

Different branch builds can be maintained by simply creating separate build
directories for different branches. When switching branches in your checkout
git will simply ignore these untracked directories.
Instead of running make clean you can simply "rm -rf build/" (although make
clean still does the usual thing when run inside build/).
You can even run "make test" within the build/ directory. This will however
symlink build/tests to the main tests/ directory and therefore created new
files within subdirectories of it (making the test system work out-of-tree
is not really feasible). You can clean this up by running (from the top
src directory!)

  git clean -df tests/

(DON'T RUN THIS IF THERE'S ANY CHANGES IN THE tests/ DIRECTORY YOU HAVEN'T COMMITTED).
You may also have to clean the python/ directory to get rid of .pyc files.


stephankramer:
Will reopen once I have some more time to look at this again and update it.

PR #19, opened 2014-08-15
Fix pvtu2vtu for DG pvtus.

stephankramer:
Previously when passing a discontinuous pvtu to pvtu2vtu, it would create a continuous vtu choosing a value at the vertex from one of the surrounding dg nodes at random. Now, it creates a discontinuous vtu without loosing any information or accuracy. Added test for new functionality.


ctjacobs:
I think this is ok.

ctjacobs:
This has been merged. Closing.

PR #20, opened 2014-08-15
Allow running project_vtu without donor mesh.

stephankramer:
Instead it works out the donor mesh from the input vtu. This only works in
serial (does project_vtu actually work in parallel?) and only for continuous
linear vtus.

New functionality switched on in project_vtu test.


ctjacobs:
Looks good to me. Ship it!

ctjacobs:
Actually, before you do that, could you also update the section on project_vtu in the manual (page 192) to show that the `donor_mesh` argument can be optional (for the continuous P1 case).

stephankramer:
Thanks, ctjacobs. Have updated the manual. Rebased branch with resolved conflicts (due to the triangle purge) is now at branch project_vtu-without-donor-mesh-rebased. Will merge that once I get green on buildbot.

PR #29, opened 2014-12-17
Changed schema for the FieldWeightedPartitionValues

stevendargaville:
This changes the adaptivity options schema to allow the FieldWeightedPartitionValues to be a diagnostic field. This lets you set spatially dependent weights for load balancing that aren't from a file/python function, etc.


stephankramer:
Looks perfectly fine

PR #31, opened 2015-01-12
Bfs reference purge

drobinson12:
The Backward Facing Step example contains two simulations: one using the k-epsilon turbulence model, and one using no turbulence model (the viscosity is raised to keep the simulation stable).  Both simulations are run by the xml and have the predicted reattachment length compared against physical data.

This example has recently failed on Buildbot and is generally known as being flaky.  I am proposing that the Reference case is the source of this flakiness and that it should be removed.  Although we might expect this simulation to solve without instability, the lack of a turbulence model (or adequate resolution to resolve turbulence) means that the simulation results inevitably bare little resemblance to the physical data to which we are comparing.  Moreover, including this in the examples suite could confuse new users who look to these simulations for guidance.


ctjacobs:
Does Stephan's recent change in `master` (bc584d86690a084daa5afee374238dcbc3251f89) fix your PETSc 3.4.2 issue? If so, you could revert 9becee7 ("Removing bug associated with PETSc 3.4.2").

tmbgreaves:
Additionally, I think I gave Dave the wrong advice about which parameter to pull out, hence this definitely needs to be reverted. Stephan's commit does solve the problem.

ctjacobs:
Great. Apart from reverting that commit, I think this is good to go. It's probably not worth getting a build queue set up since this only changes stuff in `examples`, but please make sure the example passes at least locally.

fmilthaler:
Dave, as discussed above, could you revert 9becee7 and make sure that the example still passes?

drobinson12:
Can you walk me through how to do this?  How does one "revert"?

fmilthaler:
Sure, you could either do a `revert` or `reset`.
* `git revert <commit>`: creates a new commit in the history that reverts the given `<commit>`
* `git reset --hard <commit>`: actually removes the given `<commit>` from the history

Thus, `git revert` is useful if you want to keep the changes of your commit in your history, or if other people are actively working on the branch. As far as I can tell, neither is the case here, so I think a `git reset --hard <commit>` is the better choice here since it completely removes the commit from the history.

Chris, Tim, do you agree?

ctjacobs:
Sure - done. Test the example and go.

PR #32, opened 2015-01-14
Tictoc calls for every timestep

ctjacobs:
Added tictoc calls to calculate and print out the time spent in each timestep. This excludes the time spent in mesh adapts (if applicable), which is already calculated using separate tictoc calls.


ctjacobs:
Buildbot queue is green: http://buildbot-ocean.ese.ic.ac.uk:8080/builders/timestep-tictoc

slmouradian:
Brilliant, ship it!

PR #33, opened 2015-01-16
Vtu2ensight

fmilthaler:
Adds a python script to convert vtu files to Ensight Gold files.

### DIFF
--- a/manual/visualisation_and_diagnostics.tex
+++ b/manual/visualisation_and_diagnostics.tex
@@ -1606,7 +1606,7 @@ where \lstinline[language = Bash]+basename+ is the basename of the (p)vtu output
 
 The output is sent to standard output.
 
-All files that match the pattern \lstinline[language = Bash]+basename_[0-9]*vtu+ are converted to EnSight Gold files, except for checkpoints, e.g.~vtu files that contain the word \textsl{checkpoint} are excluded from the conversion. If you want to convert \textsl{checkpointed} vtu files, use \lstinline[language = Bash]+rename_checkpoin+ (see section \ref{sec:rename_checkpoint}) first and ensure that the word \textsl{checkpoint} is removed from the basename of the solution files.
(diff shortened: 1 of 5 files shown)


manual/visualisation_and_diagnostics.tex
```diff
@@ -1582,6 +1583,31 @@ \subsubsection{vtu\_bins}
...
+ -h, --help     displays help
+ -s, --static   use this flag only when a fixed mesh was used,
+                by default a dynamically varying (adaptive) spatial mesh
+                is assumed.
+ -v, --verbose  verbose output
+\end{lstlisting}
+
+The output is sent to standard output.
+
+All files that match the pattern \lstinline[language = Bash]+basename_[0-9]*vtu+ are converted to EnSight Gold files, except for checkpoints, e.g.~vtu files that contain the word \textsl{checkpoint} are excluded from the conversion. If you want to convert \textsl{checkpointed} vtu files, use \lstinline[language = Bash]+rename_checkpoin+ (see section \ref{sec:rename_checkpoint}) first and ensure that the word \textsl{checkpoint} is removed from the basename of the solution files.
```
ctjacobs: Typo: "rename_checkpoin".

ctjacobs:
Apart from the typo this looks fine. Tim, can you set up a build queue for this branch please?

tmbgreaves:
Set up and running:

http://buildbot.ese.ic.ac.uk:8080/builders/vtu2ensight

stephankramer:
Is there anyway you can test this? Even if it's just a "does it give the same output with the same input" kind of test? We already have too many tools, that are not tested in any shape of form...

fmilthaler:
Yes, we could include such a test, however the converted output files are binaries. I wasn't sure about committing those. Is everybody ok with adding binaries for this test?

stephankramer:
Hm, in the documentation it says: "Binary files written on one system may not be readable on other
systems." So maybe this won't work. Actually looking a bit closer at this, can you not use the vtkEnsight6Reader() and do some basic tests on it?

fmilthaler:
Good point. Of course, that should do it.

fmilthaler:
Added a (short) test, which passes on buildbot. Only test failing on buildbot is internal_wave.xml (in medium tests). The same failure happened in build 227 on gcc4-x86_64.

tmbgreaves:
The failure cropped up after a system rather than code update - I have referred it to Ben but if anyone else can see a quick fix please push!

stephankramer:
Great, thanks a lot for that - looks all good to me now. The internal_wave test was fixed in 69597db737a

PR #34, opened 2015-02-06
Vtu2ensight

fmilthaler:
Adding two options to the vtu2ensight script
     * setting the (p)vtu dump number that is converted to ensight gold
     * automatically finding the last vtu dump
and removing the cellarray "vtkGhostLevels" prior to the conversion as this field caused some issues with some readers.
Manual and test are updated.


fmilthaler:
Buildbot queue is running.

ctjacobs:
Fine - go.

PR #37, opened 2015-02-17
Include prescribed fields in checkpoint

drobinson12:
One-line-change to add prescribed fields to checkpoints

### DIFF
--- a/femtools/Checkpoint.F90
+++ b/femtools/Checkpoint.F90
@@ -626,7 +626,8 @@ contains
               & .or. (have_option(trim(s_field%option_path) // "/prescribed") &
                 & .and. interpolate_field(s_field))) &
               & .and. .not. aliased(s_field) &
-              & .or. have_option(trim(s_field%option_path) // "/diagnostic/output/checkpoint") ) then
+              & .or. have_option(trim(s_field%option_path) // "/diagnostic/output/checkpoint") &
+		& .or. have_option(trim(s_field%option_path) // "/prescribed/output/checkpoint") ) then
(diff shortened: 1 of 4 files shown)


drobinson12:
I have run make test on my local computer and this branch passed.  Tim, do you want to create a Buildbot queue, or are you happy for this to be shipped?

tmbgreaves:
Assuming nobody else objects I'm happy for this to go straight in.

cianwilson:
No objections to this but curious why it is required/desired?

cianwilson:
See now that the option was already in the schema (presumably added accidentally when time averaged diagnostic fields were included) and that this just adds the corresponding functionality to the code.  Still curious why it's useful but assuming it is could the schema documentation (which for prescribed fields still talks about time-averaged diagnostic fields) be updated please?

drobinson12:
In this instance, I'm trying to flredecomp a simulation that prescribes 
a field using a vtu.  As it stood, the code didn't output a partitioned 
version of the input vtu.

stephankramer:
Ah ok, that actually points to a slightly different bug: if you have a prescribed field with the from_file option it should flredecomp that correctly with or without the output/checkpoint option. Actually, you can already get the correct behaviour if you add interpolation options to the prescribed field. If you use adaptivity you will need these anyway, because after an adapt it can't re-prescribe it from a python function as usual, but needs to interpolate from the previous mesh.

In any case it should also adhere to output/checkpoint option. So I'm happy to have this change in - if you could also fix the documentation as Cian requested.

fmilthaler:
Thanks for fixing the documentation in diamon as Cian suggested. This looks good now. Ship it.

PR #41, opened 2015-03-30
Analytical sphere rebased

stephankramer:
This branch implements a new approach to the idea of bended (or curved) elements in spherical geometries. Like before, the embedding of elements in physical space is curved by moving points in the interior (and along edges) in the radial direction such that their radial coordinate is interpolated between the vertices of the triangle or tet. In particular this means that a triangle with all its nodes on a perfect sphere, is curved in such a way that it lies entirely on the sphere. Unlike the existing superparametric_mapping option however, where the curved triangles and tets are approximated with a higher order polynomial by using a higher order CoordinateMesh, the new analytical_mapping does not require a higher order CoordinateMesh. Instead it applies the exact analytical formula for the bending of the element in its interior and computes an analytical expression for the jacobian on the gauss points (the only place where that is actually affected by the bending). Thus the approximation of the bended element is only limited by the degree of quadrature.

Thorough testing of this branch (thanks rhodrid) revealed an important issue with the computation of gradients on surfaces. The issue with the computation of a gradient on a facet is that it relies on all the values in its adjacent volume element, not just the values on the facet itself. For this purpose the femtools code has the concept of an augmented shape function which allows you to compute the gradients of all shape functions in a volume element on the gauss points of all its adjacent facets. The problem with the augmented shape functions is that it seems to only rely on the ordering of the volume element, since we only multiply it with nodal values of the entire element (ele_val(...)). This is misleading because it returns gradients on gauss points on the facets, and these gauss points come out in a certain order. This order is determined by the canonical ordering of facets around a volume element. This is not the same ordering for facets in a femtools mesh (as returned by e.g. face_val(), face_global_nodes(), etc.) which is chosen in a (semily) arbitrary way to ensure the ordering of the two opposite facets between two elements have the same ordering. This has not been an issue before because the gradients were only multiplied by things based on the Jacobian which for flat elements is constant over the gauss points. With bended elements, the order of the jacobian computations over the gauss points does matter however.

To fix this we (that is, Cian really) have changed the way augmented shapes are used in femtools/fluidity:
- augmented shapes are now always added to the mesh%shape whenever add_faces() is called
- new functions face_n_s() and face_dn_s() return the shape functions and gradients of the shape functions on the gauss points of a facet in an order that is consistent with the facet node numbering of a given mesh. For this reason the function needs the global element ad facet number and the mesh object.
- transform_to_physical_facet() is now also able to return the dshape (i.e. the gradient transformed to physical space of a given shape function (not necessarily the X shape)), and the jacobian and inverse jacobian all at the gauss points in and order that is consistent with the ordering of the given global facet number.
These changes mean that the whole rigmarole of checking whether a shape has augmented shape functions, if not allocating a local shape function that does have them, then doing complicated stuff with jacobians - all of which was copied and pasted in a few places in fluidity is now all wrapped in a single call to transform_to_face_physical().

Other changes in the code:
- compute_jacobian() and compute_inverse_jacobian() now need the entire coordinate field and an element number instead of just providing ele_val(X, ele). This is to ensure we can check the settings of the coordinate field with regards to whether it should be bend or not. This is also consistent with all the other transform_to_physical functions.
- superparametric_mapping and analytical_mapping are now independent on/off switches under /geometry/spherical_earth. This is because it still makes sense to apply the analytical mapping to a coordinate mesh that is already superparametrically curved. In particular if higher order (p>1) fields are calculated on the nodes (e.g. prescribing it from a python function), it may be necessary to have the correct curved coordinates available on the "higher order nodes". This means superparametric_mapping is no longer a a choice between linear_mapping and superparametric_mapping. This means that existing flmls with spherical_earth will loose their linear_mapping option (the default) if they had them. There should be no compatibility option of running an older flml under the new code however.
- with /geometry/spherical_earth read_gmsh no longer assumes the positions%dim==3 for the input mesh but instead takes the positions%dim value from /geometry/dimension/ (the topological dimension may be lower when using extrusion with spherical earth). This is to enable circular domains (cylindrical) extruded from a circle in two/spherical_mappings_2d dimensions. This is simply a 1-sphere embedded in 2 dimensions, so there's no reason spherical_earth should not support it.
- in vertical integration, fix a nloc==nvertices assumption

The description /geometry/spherical_earth has been updated in the manual.

Tests
====================================
- new tests spherical_mappings_2d and spherical_mappings_3d test the new analytical mapping on volume integrals and surface integrals in 2d an 3d and compares them with the default (no bending) and superparametric. For analytical_mapping it seems that we can indeed get arbitrarily (limited by quadrature) close to the analytical answer on a very coarse mesh, so that gives you an interesting way of computing pi by calculating the area of a unit circle.
- new test analytical-mapping-tracer-cg (courtesty Alex Avdis) tests this functionality for ocean modelling purposes. It runs a linear stratified ocean on a very coarse mesh of a small part of the sphere ("spherical patch") with P1DG-P2, temperature is P1CG. This would not be stable for very long without analytical_mapping, and is only a little more stable with P2 superparametric. With analytical mapping it remains stable for much longer.


jrper:
After sitting down with Stephan for him to explain which transformation is actually being applied, I'm happy for this to go on green with a paragraph added to the manual. 

I guess it would be a nice extension to allow for user defined metrics as well, but that's definitely beyond the scope of these changes.

stephankramer:
Thanks for the review James. This has now been merged.

cianwilson:
Many thanks James!

PR #42, opened 2015-04-07
Reorder the mesh formats in the schema so that gmsh is the default format.

cianwilson:


stephankramer:
Looks eminently sensible.

PR #43, opened 2015-04-10
Shared object initialisation

jrper:
Branch implements field initialisation from code in external shared objects. This is faster than python initialisation, but the principle benefit is the capability to link in more general external codes via the same method.

### DIFF
--- a/preprocessor/dl_fortran.F90
+++ b/preprocessor/dl_fortran.F90
@@ -1,0 +1,47 @@
+!! file dl_fortran.F90
+
+!! wrapper for c dlopen functionality.
+!! based on examples at http://cims.nyu.edu/~donev/Fortran/DLL/DLL.Fortran.txt
+
+MODULE DL_fortran
+   USE, intrinsic::  ISO_C_BINDING
+!   USE ISO_C_UTILITIES
+   IMPLICIT NONE
+   PRIVATE
+
+   PUBLIC :: DLOpen, DLSym, DLClose, DLError ! DL API
+   
+   ! Valid modes for mode in DLOpen:
+   INTEGER, PARAMETER, PUBLIC :: RTLD_LAZY=1, RTLD_NOW=2, RTLD_GLOBAL=256, RTLD_LOCAL=0
+      ! Obtained from the output of the previously listed C program 
+         
+   INTERFACE ! All we need is interfaces for the prototypes in <dlfcn.h>
+      FUNCTION DLOpen(file,mode) RESULT(handle) BIND(C,NAME='dlopen')
+         ! void *dlopen(const char *file, int mode);
+        USE, intrinsic::  ISO_C_BINDING
+         CHARACTER(C_CHAR), DIMENSION(*), INTENT(IN) :: file
+            ! C strings should be declared as character arrays
+         INTEGER(C_INT), VALUE :: mode
+         TYPE(C_PTR) :: handle
+      END FUNCTION
+      FUNCTION DLSym(handle,name) RESULT(funptr) BIND(C,NAME='dlsym')
+         ! void *dlsym(void *handle, const char *name);
+        USE, intrinsic::  ISO_C_BINDING
+         TYPE(C_PTR), VALUE :: handle
+         CHARACTER(C_CHAR), DIMENSION(*), INTENT(IN) :: name
+         TYPE(C_FUNPTR) :: funptr ! A function pointer
+      END FUNCTION
+      FUNCTION DLClose(handle) RESULT(status) BIND(C,NAME='dlclose')
+         ! int dlclose(void *handle);
+        USE, intrinsic::  ISO_C_BINDING
+         TYPE(C_PTR), VALUE :: handle
+         INTEGER(C_INT) :: status
+      END FUNCTION
+      FUNCTION DLError() RESULT(error) BIND(C,NAME='dlerror')
+         ! char *dlerror(void);
+        USE, intrinsic::  ISO_C_BINDING
+         TYPE(C_PTR) :: error
+      END FUNCTION         
+   END INTERFACE
+      
+END MODULE
--- a/preprocessor/plugin_functions.F90
+++ b/preprocessor/plugin_functions.F90
@@ -1,0 +1,136 @@
+!    Copyright (C) 2007 Imperial College London and others.
+!    
+!    Please see the AUTHORS file in the main source directory for a full list
+!    of copyright holders.
+!
+!    Prof. C Pain
+!    Applied Modelling and Computation Group
+!    Department of Earth Science and Engineering
+!    Imperial College London
+!
+!    amcgsoftware@imperial.ac.uk
+!    
+!    This library is free software; you can redistribute it and/or
+!    modify it under the terms of the GNU Lesser General Public
+!    License as published by the Free Software Foundation,
+!    version 2.1 of the License.
+!
+!    This library is distributed in the hope that it will be useful,
+!    but WITHOUT ANY WARRANTY; without even the implied warranty of
+!    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+!    Lesser General Public License for more details.
+!
+!    You should have received a copy of the GNU Lesser General Public
+!    License along with this library; if not, write to the Free Software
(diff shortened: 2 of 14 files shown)


tmbgreaves:
The build and test side looks fine to me; I'm no Fortran expert, though, so whilst this also looks fine I'd defer to someone else who knows more about Fortran and Fluidity. One comment - the URL you give as the source for the dl_fortran code returns 404; I think the correct URL should be http://cims.nyu.edu/~donev/Fortran/DLL/DLL.Forum.txt

I *think* in most cases that dlopen should come directly from libc in GNU operating systems at least, but it can't hurt to have a check for this.

I've set up a buildbot queue for the branch: http://buildbot.ese.ic.ac.uk:8080/builders/shared_object_initialisation

preprocessor/plugin_functions.F90
```diff
@@ -0,0 +1,136 @@
...
+       end if
+       ! we've just allowed remapping from a higher order to a lower order continuous field as this should be valid for
+       ! coordinates
+       ! also allowed to remap from unperiodic to periodic... hopefully the python function used will also be periodic!
+
+       fposition => lposition
+    end if
+
+
+    handle=DLOpen(plugin_name//C_NULL_CHAR,IOR(RTLD_NOW,RTLD_GLOBAL))
```
stephankramer: I would stick a trim() around the plugin_name there, just in case someone calls this with some trailing spaces. Could you add a comment what these flags do?

preprocessor/plugin_functions.F90
```diff
@@ -0,0 +1,136 @@
...
+  procedure(val), pointer ::dll_sub
+
+
+
+  if (field%mesh==position%mesh) then
+
+     fposition => position
+  else
+       ! Remap position first.
+       call allocate(lposition, position%dim, field%mesh, "Local Position")
```
stephankramer: The indentation is wonky.

preprocessor/plugin_functions.F90
```diff
@@ -0,0 +1,136 @@
...
+    handle=DLOpen(plugin_name//C_NULL_CHAR,IOR(RTLD_NOW,RTLD_GLOBAL))
+
+    if (.not. c_associated(handle)) then
+       call c_f_pointer(DLError(),fstring)
+       slen=index(fstring,c_null_char) - 1
+       FLAbort( 'error in dlopen:'//fstring(1:slen) )
+    end if
+
+
+  funptr= DLSym(handle,function_name//'_'//C_NULL_CHAR)
```
stephankramer: Again, you could trim() it just to be safe...

preprocessor/plugin_functions.F90
```diff
@@ -0,0 +1,136 @@
...
+    type(C_PTR) :: handle=C_NULL_PTR
+    type(C_FUNPTR) :: funptr=C_NULL_FUNPTR
+
+    integer(C_INT) :: status
+    type( vector_field ), target :: lposition
+    type( vector_field ), pointer :: fposition
+
+    integer :: stat
+
+    abstract interface
```
stephankramer: I might be wrong, but I don't think this needs to be an abstract interface. In general we don't (yet) depend on the OO features of F2003 in Fluidity, and I'm not sure they have been fully implemented in all compilers we use.

preprocessor/dl_fortran.F90
```diff
@@ -0,0 +1,47 @@
+!! file dl_fortran.F90
+
+!! wrapper for c dlopen functionality.
+!! based on examples at http://cims.nyu.edu/~donev/Fortran/DLL/DLL.Fortran.txt
+
+MODULE DL_fortran
```
stephankramer: What's with the CAPS?

tests/initialise_from_shared_object/Makefile
```diff
@@ -0,0 +1,9 @@
+default: input
+
+input: clean
+	gmsh -2 src/box_2d.geo
+
+	gfortran -fPIC -shared -g -o example.so src/example.F90
```
stephankramer: Hm, I guess we need to think of something more general here. We can't rely on gfortran being available.

stephankramer:
I like the idea in general. I've added some comments here and there.

One though I had - up for discussion: it currently tells people to use an implicit interface fortran function, whereas we could tell people to provide a functino that uses the C ABI (i.e. uses bind(c)).
Advantages:
* we no longer rely on the fortran-name-in-small-caps-plus-underscore convention which is followed by the compilers we use, but is not really guaranteed.
* Also I'm not entirely sure assumed shape arrays, dimension(:), are guaranteed to work with implicit interfaces. I thought that wasn't allowed at all. It is certainly not guaranteed to work across compilers (i.e. if fluidity is compiled with a different compiler than the shared library).
* people could write their function in C or C++ (with extern "C")
Disadvantages:
* you'd have to think about the arrays, probably passing in the size separately as per C convention, and make assumptions about row vs. column major order.

stephankramer:
Just to clarify my point about assumed shape arrays above, as I slightly mixed up terminology here: explicit vs. implicit is about the way you call a subroutine, not how it's defined. The way you call the val subroutine currently, it has an *explicit* interface (as you provide an interface block). It is true that assumed shape is only allowed with explicit interface.

However this makes the point I was trying to make only worse: the old fashioned way of calling a fortran routine in an interoperable way (e.g. from C) was by defining it in such a way that it is *possible* to be called with an implicit interface (i.e. outside a module, no fancy array types), as this prevents the compiler from doing all sorts of internal magic about how things are passed onto the stack and typically results in arrays just being passed in the C-way. Having an implicit interface for the call with assumed shape arrays means the compiler of the calling code will make all sorts of assumptions about for instance how the stride is passed onto the subroutine and these assumptions will break down even between different versions of the same compiler. This makes the whole point of compile fluidity only once, and only recompile the shared lib a bit dubious.

You could fix it by having an interface like:
   subroutine val(n, dim, field, x, t)
      integer n, dim
      real, dimension(n):: field
      real, dimension(dim,n):: x
      real t
  end subroutine
but then again, we might as well go the whole way and make it bind(C).

drhodrid:
Just a quick comment: could this branch be tested on the intel build prior to a trunk merge? Just to verify that it doesn't break that?

tmbgreaves:
It's not very easy to test it pre-merge with the Intel buildbot, unfortunately - the way it's set up we can't easily replicate the testing framework a second time for a branch.

jrper:
I've attempted to have a first pass at addressing @stephankramer's criticisms of the dynamic loading side of things, a bit more work to follow shortly once I've finished testing on CX1. It appears that the use of the abstract statement is required to produce an interface that doesn't automatically bind by name, but there may still be some magic which could be done to avoid that.

stephankramer:
Hm, the original version (by @jrper) looks correct to me, whereas I'm not sure that @drhodrid 's is (#ifdef should only be followed by the symbol to be tested and not a logical expression).

@drhodrid: could you send us a config.log of what happens if you try to configure parmetis 4 with @jrpers's version?

stephankramer:
Maybe I should clarify myself a little more: I never suggested people to write their initialisation in C (although it's nice that people can now), rather I suggested we should tell people to write their *fortran* initialisation routine using bind(C) so that it uses the C ABI. This is the only way to guarantee compatibility between different fortran compilers (and versions) as there is no such thing as a fortran ABI.

Moreover, since this implementation already makes heavy use of bind(c), there is no reason why people can't use it for themselves. As a side effect it would simplify things as we'd only have to support the "C" interface_type. I just tried the following subroutine called from fluidity using the "C" interface_type in your branch and it seems to work fine:

        subroutine val(dim,n,field,X,t) bind(C)
            use iso_c_binding
            integer(c_int) :: dim, n
            real(c_double), dimension(n) :: field
            real(c_double), dimension(dim,n) :: X
            real(c_double) :: t

            field=X(1,:)**2-X(2,:)

        end subroutine val

tmbgreaves:
Once more following up from today's dev meeting - bonus Karma to Stephan for agreeing to have a glance through this and merge if he's happy or feed back to James if not :-)

tmbgreaves:
Follow up here - Stephan, I'm guessing that the additional commits above were a result of comments from you; do you have any further thoughts on this one and/or can it be merged?

PR #44, opened 2015-04-10
Make radial gravity at gauss points an option rather than automagic 

cianwilson:
Currently when a user turns on the /geometry/spherical_earth option a number of things happen automagically, with no user control.  This branch implements the first fix for this by separating out the automatic placing of radial gravity at gauss points into its own option underneath prognostic velocities.  This explicitly states what the option will do (i.e. ignore your gravity direction for the buoyancy terms) and allows the functionality to be tested properly and independently (of the various other things that happen automatically).

A test has been added on a circle and hemisphere to check that the hydrostatic pressure is recovered when solving using P1DGP2 and P2P1. 

Green buildbot branch here: http://buildbot-ocean.ese.ic.ac.uk:8080/waterfall?show=spherical_gravity_at_gauss_point_option

### DIFF
--- a/manual/configuring_fluidity.tex
+++ b/manual/configuring_fluidity.tex
@@ -133,32 +133,25 @@ The quadrature degree is specified by setting \option{/geometry/quadrature/degre
 \index{mesh!superparametric}
 
 Enabling \option{/geometry/spherical\_earth} informs \fluidity that your
-simulation is being carried out in an Earth like geometry. In three dimensions
-this means that the mesh domain is assumed to approximate (a part of) a sphere, but this option can
-also be used in two dimensions where the domain is assumed to be (a part of) a
-circle (cylindrical domain). The domain is still described in
-general $x,y,z$ coordinates (no spherical coordinates)
-and the centre of the N-sphere must coincide with the origin. Gravity is assumed
-to point to the centre. Generally, the sphere is only extruded to a certain
+simulation is being carried out in an Earth like geometry. This means that the mesh domain is assumed to approximate (a part of) an
+N-sphere (sphere in three dimensions, circle in two dimensions). The domain is still described in
+general $x,y,z$ coordinates (no spherical/cylindrical coordinates)
+and the centre of the N-sphere must coincide with the origin.  Turning on this option has a number of effects.
+
+In combination with extrusion options (see section \ref{sec:extruded_meshes}), 
+specifying a \option{spherical\_earth} will extrude radially from the top surface of the
+domain towards the origin.  Generally, the sphere is only extruded to a certain
 depth, leaving a void around the origin. The two main applications of this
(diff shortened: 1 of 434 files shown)


manual/configuring_fluidity.tex
```diff
@@ -133,32 +133,25 @@ \subsubsection{Spherical Earth}
...
-Another consequence of this option is that the direction of gravity is no longer
-read from the \option{GravityDirection} field (under
-\option{/physical\_parameters/gravity}), but is always assumed to be the
-negative radial direction and is directly calculated at the gauss points
-(instead of being interpolated from the mesh nodes). It is however still advised
-to setup the \option{GravityDirection} from a python function that returns
-the same negative radial unit vector, as some places in the code still expect
-it.
+In general, when using this option you will want to specify the gravity direction to be opposed to the radial
+direction (i.e. pointing to the origin) however it can also be used in conjunction with the
```
stephankramer: This sentence is a little confusing (in particular the "however") and long.

stephankramer:
This looks great! Thanks for adding the warning. This has also been discussed off-line with people that may be affected by this for their ocean on the sphere runs. So other than my minor comment above, I'd say good to go.

cianwilson:
Thanks Stephan!

stevendargaville:
Just wondering if anyone has seen any issues in this commit with both OpenMP and debugging enabled?

We have some code that is calling construct_advection_diffusion_dg, and it is failing an assert on line 1017 of Advection_DIffusion_DG.F90. It seems the cache is invalid?

cianwilson:
I'm dubious that this merge could cause that error as it has nothing to do with either the cache or dg advection.

Do you have an example flml that demonstrates the problem?

stevendargaville:
So I'm getitng failures in the following tests with both debugging and OpenMP enabled:

mmat-impact.xml: F
1mat-shocktube-gmsh.xml: F
tracer_inflow_2d_dg_cg_quad.xml: F
mmat-gravity-col-p2lumpedp1.xml: F
gauss_point_buoyancy.xml: F
1mat-shocktube.xml: F
mmat-shocktube.xml: F
spherical_patch.xml: F
mmat-gravity-col-p2p1.xml: F

So these failures are all for the same reason (invalid cache), but are not limited to advection, some of these fail in Momentum_DG.F90. Looking at the cache invalidation (prepopulate_transform_cache in Transform_Element.F90), all that is required to trigger it is a different refcount in the coordinate mesh, compared to when the caching occured.

stevendargaville:
Terribly sorry, turns out it was not this commit that caused the issues, these errors starteds appearing after a different commit on the 25th of Feb (2382bc2b6be13c6d78ce59cc919a1a3de42180d4). I'll raise an issue for it.

PR #45, opened 2015-04-22
Possible fix for prefix not getting expanded in flml schemata file on install?

cianwilson:
Not so much a pull request as a request for better suggestions/opinions.

When I configure with `--prefix=...` and make install I'm finding that the schemata file in `prefix/share/diamond/schemata/flml` is invalid because the path listed in that file still has `${prefix}` in it (to see this just run configure then look at `schemas/flml`).  This is because `schemas/flml.in` uses `@datadir@` which is not expanded out beyond `${prefix}/share`.  There is a comment in configure that suggests this is done intentionally so that users can run `make install exec_prefix=...` but that workflow wouldn't work for this file as far as I can tell as install just copies the file into the destination directory without further (post configure) modifications.

So, I'm wondering if anyone uses or is aware of the intentions of that functionality?  If not, can we just remove references to `datadir` and have configure expand `$prefix`.

In the meantime this diff does that for `schemas/flml.in`, which gets me around this problem for now.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -491,7 +491,7 @@ distclean: clean
 	 include/spatialindex include/spud include/spud.h		\
 	 include/spud_enums.h include/tinystr.h include/tinyxml.h	\
 	 include/version.h include/vtk.h				\
-	 preprocessor/check_options.F90 schemas/flml			\
+	 preprocessor/check_options.F90 \
 	 preprocessor/register_diagnostics.F90 python/setup.py >	\
 	 /dev/null
 	@for i in `find ./*/ -name Makefile.in`; do rm -f `echo $$i | sed 's/.in$$//'`; done > /dev/null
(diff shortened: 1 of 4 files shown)


stephankramer:
So after quickly discussing this with Cian and some more googling, I think I changed my mind. It seems that the autoconf documentation very clearly specifies that prefix is not necessarily fixed at configure time and people should be able to do:

    make install prefix=<path-that-is-different-than-the-prefix-specified-at-configure>

If we want to follow that (and there's no real reason we can't) the fix here doesn't work either, because schemas/flml was created at configure time so it would still point to the old prefix. This could be easily fixed by removing schemas/flml from the list of files to be processed by configure, and create schemas/flml from the Makefile by a simple:

    sed 's/@datadir@/$datadir/g' schemas/flml.in > schemas/flml

cianwilson:
OK, new attempt at a fix implemented using sed as suggested.

cianwilson:
Any further comments on this?  (Was just installing fluidity and hit this bug again.)

tmbgreaves:
Sorry, this had dropped off my radar. I've done a bit of playing around and can't think of anything neater than this, and like this solution.

cianwilson:
Many thanks Tim,  I've rebased (and squashed) this and completed the merge.

Issue #46, opened 2015-04-24
Test failures with debugging and OpenMP


stevendargaville:
There appears to be a problem (stemming from commit 2382bc2b6be13c6d78ce59cc919a1a3de42180d4) when using both OpenMP and debugging. I'm getitng failures in the following tests

Summary of test problems with failures or warnings:
mmat-impact.xml: F
1mat-shocktube-gmsh.xml: F
tracer_inflow_2d_dg_cg_quad.xml: F
mmat-gravity-col-p2lumpedp1.xml: F
gauss_point_buoyancy.xml: F
1mat-shocktube.xml: F
mmat-shocktube.xml: F
spherical_patch.xml: F
mmat-gravity-col-p2p1.xml: F

So these failures are all for the same reason (invalid cache), but are not limited to advection, some of these fail in Momentum_DG.F90. Looking at the cache invalidation (prepopulate_transform_cache in Transform_Element.F90), all that is required to trigger it is a different refcount in the coordinate mesh, compared to when the caching occured.

This wasn't picked up as there is no combined OpenMP and debugging build on buildbot, I'll email Tim to see if he can add one.

cianwilson:
The thing that all those tests have in common is that they shouldn't be using the cache at all (some are cube elemtents, others are and/on a sphere or have a higher order coordinate mesh).  prepopulate_transform_cache was changed in @563127720ad354b61441aafdfa4879dcef362b7c to accurately return whether the cache was valid or not (previously it always returned true).

I'm unclear why it does return a flag or even checks it.  As far as I can tell nothing bad will happen if you proceed from this point since the cache will never be valid and will therefore never be constructed (which appears to be the problem with openmp) or used.

So, I think that assert just needs to be removed...  Any thoughts from anyone familiar with openmp?

wence-:
So when I wrote that bit of the OpenMP code, I couldn't think hard enough about how to make the cache population thread safe (hence adding the original aborts).  It's been a while since I looked at any of that code, but the question is, I guess, why the cache population is being called?

cianwilson:
This is in prepopulation of the cache, which attempts to construct the cache outside of the threaded bit of the code.  However as using the cache is invalid, cache construction isn't actually being called.  It's just returning a flag to let you know that.

wence-:
Oh yes, I see.  Yeah, that assertion can almost certainly be removed.

cianwilson:
@stevendargaville : Could you give it another go now please?

[...]

PR #47, opened 2015-04-24
Remove the asserts following cache prepopulation with openmp as there's no reason they can't be false.

cianwilson:
This is a possible fix for issue #46 .


wence-:
Yes, that looks right, although I'm dredging from somewhere in the depths of memory.

PR #48, opened 2015-04-28
Petsc master fixes

stephankramer:
Make fluidity build with petsc master again

This is to cope with new layout of petsc installation build from master. Changes:
* include/finclude -> include/petsc/finclude.

This means that each fortran file should have:

    #include "petsc/finclude/include_file.h"

instead of

    #include "finclude/include_file.h"

* conf -> lib/petsc/conf. This means the included makefiles in the petsc_makefile have to change

To avoid having to support both old and new conventions with lots of `#ifdef`s, this is all done in the petsc_legacy.h header file, which is now used everywhere including all the tools.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -428,9 +428,10 @@ AC_MSG_NOTICE([Using PETSC_DIR=$PETSC_DIR])
 PETSC_LINK_LIBS=`make -s -f petsc_makefile getlinklibs`
 LIBS="$PETSC_LINK_LIBS $LIBS"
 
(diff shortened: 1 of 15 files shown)


drhodrid:
This is one of the most beautiful things I've ever seen. Looks good to me. It has a green buildbot and I ran a local make test using petsc master where everything passed. GORGEOUS! Get it in ;-)

mlange05:
Yes, looks good to me too. Thanks Stephan

PR #51, opened 2015-05-08
Changes to build with parmetis 3 or 4 with autoconf testing for version no.

jrper:
This adds the magic preprocessor directive -DPARMETIS_V3.

Rhodri, over to you to test this under as many setups as possible.


drhodrid:
Percy - I had a look at this today. It was quite tough to make fluidity interact solely with the new version of parmetis but think I figured it out in the end. I think the following change is needed in configure.in:

 #include <parmetis.h>
-#if defined(PARMETIS_MAJOR_VERSION) && PARMETIS_MAJOR_VERSION <= 3
+#ifdef (PARMETIS_MAJOR_VERSION) && PARMETIS_MAJOR_VERSION <= 3
 test_true
 #endif

Or else the macro gets picked up even when building with parmetis 4. Maybe somebody better qualified can verify this?

With parmetis 4 and your changes, fldecomp works fine and I have a full working short test suite - so you seem to have got the API right. As I've mangled my local libraires (!) I cannot currently test this with parmetis version 3... So, my recommendations:

1. Incorporate the above change and run autoconf. 
2. Set up a buildbot branch (which has the older versions of metis and permetis) to verify that old functionality remains.
3. Get somebody that better understands autoconf to glance through your changes.

One other issue - do we need to make makefiles? I never can tell when this is required!

Anyway - apart from the above... I think it's good to go. Nice work.

jrper:
Thanks Rhodri. I think your version of the C preprocessor directive is doing something slightly different from what I intended, but I'll try and find someone (@stephankramer maybe?) who understands that double sorcery better to fix it up. Autoconf is a silly language. I'll also ask tim for a buildbot queue.

I don't *think* make makefiles is strictly necessary, since none of the dependencies have actually changed, regardless of preprocessor directives, but if I run it anyway, it should be safe.

stephankramer:
Hm, the original version (by @jrper) looks correct to me, whereas I'm not sure that @drhodrid 's is (#ifdef should only be followed by the symbol to be tested and not a logical expression).

@drhodrid: could you send us a config.log of what happens if you try to configure parmetis 4 with @jrpers's version?

tmbgreaves:
Branch queue at: http://buildbot.ese.ic.ac.uk:8080/builders/parmetis_v4

jrper:
Post offline discussion with @stephankramer and @tmbgreaves, I propose the following (see commit 812097d7619835f084033063f98727a8ffd04e8b): PETSc is now searched for before ParMETIS, METIS and Zoltan, and if a simple program links, no further libraries are added, otherwise the existing automatic routines are called. 

@drhodrid Is there any chance you could try this on your system? Set $PETSC_DIR as required before configuring, but don't add additional flags.

stephankramer:
Moving this back online: may I suggest a slightly different fix (for the fact that fldecomp doesn't pick up PETSc's include flags and therefore fails to find a parmetis build as part of PETSc). The reason it's not picked up is because the build rule in fldecomp/Makefile.in for .cpp files does not use CPPFLAGS - whereas I think it should (at the very least to be consistent with what we do in other places). If you stick the petsc include flags in CXXFLAGS as well, then in all other places (where we do use both CPPFLAGS and CXXFLAGS) you will get the whole thing twice on the command line and the petsc include flags can be quite a long expression already...

stephankramer:
Another iteration of this. @drhodrid: could you try again please with your petsc with parmetis4 build (without specifying any additional flags) and run the tests?

drhodrid:
Hi Stephan,

That works fine for me. It is no longer picking up the old parmetis (again, as far as I can tell!), and all tests pass.

I have attached the config.log to this email in case you want to take a quick peek.

On a (at least in my eyes!) related note - will the changes made to configure also ensure that fluidity picks up any version of zoltan installed through petsc, before looking elsewhere on the system? It so, great stuff. If not, could configure be further tweeked to ensure that this happens? If its the latter, I’m happy to open another discussion thread on this…

R

jrper:
This has a green buildbot, and Stephan seems happy now, so I'll squash the changes and push to master.

@drhodrid This doesn't interact with the zoltan configuration directly, but ought to ensure that any zoltan libraries in $PETSC_DIR take precedence over system zoltan. However finding the .mod files may cause some problems. I suggest opening a new thread and homely somebody will have enough time/coffee to fix it.

PR #52, opened 2015-05-18
Sndgln fixes

stephankramer:
Various fixes to increase robustness of internal boundaries.

In the default way of calling add_faces (without element_owner),
the surface mesh sndgln should only contain each internal facet
once. This is now checked for and an error message is produced, 
for instance when the user has assigned a physical id to parts of the 
surface (internal or external) more than once in the .msh input file.

The call to add_faces *with* element_owner argument (used for periodic
meshes) otoh expects interior facets to appear twice in sndgln (possibly
with different boundar id). This is to support two different surface ids
on each side of an internal facet which is supported to handle periodic
meshes only. This should therefore not have been used in  zoltan (and
its renumbering routines) by default as it will cause the ordering of
interior facets to be broken and extra element ownership to be written
upon checkpoint. To help zoltan remove duplicate interior facets (for
the case without element ownership), add_faces is now able to remove
these for you (if you kindly ask it). In this case, it will mean the surface mesh
stored on the mesh object is no longer numbered in the same way as
sndgln. It will still check that duplicate facets have the same facet
id.

To be more consistent with the above, getsndgln will now return the
first unique_surface_element_count() facets only, i.e. giving each
external and interior facet once only. For the case without element
ownership, this is the same number of facets originally provided
(unless you've asked add_faces() to remove duplicates in the input).
With element ownership, unique_surface_element_count() is the same as
surface_element_count() so nothing changes. Since this sndgln is
typically used in adaptivity wrappers where it is paired with
interleaved boundary+coplanar ids, the relevant routines in
surface_id_interleaving also return these ids for the first
unique_surface_element_count() facets only. Boundary ids and
coplanar ids should be consistent on internal facets, and we only want
to provide them once to adaptivity for locking. Meshes with element
ownership (only used for periodic meshes) should never reach the 
adaptivity layers.

This branch also fixes the handling of internal boundaries in combination 
with prescribed regions, where the internal boundary is internal 
to the non-prescribed region in particular in parallel. This is now tested in 
the internal_boundary_adaptivity_parallel test.


cianwilson:
Looks good to me.

PR #53, opened 2015-05-18
Edited the logic within pvtu2vtu.py so that it accepts [pvtu name] as…

drobinson12:
… the single input argument.


fmilthaler:
could you please update the manual as well?

fmilthaler:
looks good now, buildbot is green too, merge it.

PR #54, opened 2015-05-18
Two  sets of changes to assemble/Sam_integration.F90

jrper:
1) Reduce a double instance of 'use detector_tools' down to one. This makes local builds on intel a lot easier.

2) add an additional (and unused) stat to remap_field calls in sam_drive. This gets round interface changes which meant that the code would fail for P2 fields.


tmbgreaves:
I've set up a builder at: http://buildbot.ese.ic.ac.uk:8080/builders/sam_fix which tests this by breaking the zoltan install, then doing a build and test without working zoltan.

stephankramer:
ad 2): Could you expand a bit on why you think it shouldn't be failing when it's fed a P2 field? In general, when a stat argument is passed its return value should always be checked and only allow for specific cases that are considered harmless. Otherwise god knows what error might be silently repressed (this might also be an error condition that is added in the future to a stat-bearing subroutine).

jrper:
Based on a shallow reading of the code, the goal of the remap_field calls is precisely to extract the linear component of fields before passing into sam, which once upon a time a call to remap_field did without any error checking at all. It now error checks, which means that when sam_drive is called, it fails. I guess the polite way to do this is to check the error code really is from a mapping from high to low order?

jrper:
Comparing http://buildbot.ese.ic.ac.uk:8080/builders/sam_fix running this branch to http://buildbot.ese.ic.ac.uk:8080/builders/trusty-sam running on master, I see 11 failures here (3 short, 8 long)  versus 22 there (11 short, 11 long)
The short test failures are:
spherical_patch.xml: (uses radial extrusion)
sigma_layer_sphere_parallel.xml: (I think uses radial extrusion)
heat_2d_dg_local_assembly.xml: (is dg)
The long test failures are:
analytical-sphere-cg-tracer.xml: (I think uses radial extrusion)
geopressure_convergence.xml: [complains about quadrature, this I don't understand]
flredecomp_2d_fieldweighted.xml: [fails to construct new mesh with array bounds error]
geopressure_convergence_p1.xml: [complains about quadrature]
viscous_fs_zhong_spatial.xml: [fails to construct new mesh with array bounds error]
wetting_and_drying_balzano3_cg_parallel.xml: [fails to construct new mesh with array bounds error]
standing_wave-parallel.xml: [fails to construct new mesh with array bounds error, extruded]
mmat-balancepressure-extrude-parallel.xml: [fails to construct new mesh with array bounds error, extruded]
so I *think* the remaining errors here are features it doesn't have, but I can't seem to find documentation on the point. Does @ggorman have an opinion?

stephankramer:
Any test with an extrude/ option or any DG field should be marked to run with zoltan only. fieldweighted also sounds like something that should only work with zoltan.

In any case, ideally every of these should end with an FLExit("Functionality X is not supported with libsam, please reconfigure with zoltan") - although for DG fields this might be tricky to achieve - for extrude it should be easy to do as an options check. An array bounds error is never an acceptable way of failing.

jrper:
Presumably dg could be options_checked by an exhaustive check of mesh_continuity options, with attendant false positives, but at least it only needs to be run once. Sam_Integration.F90 already options checks for hr_adaptivity/preserve_mesh_regions (and gives the slightly unhelpful message that they don't work in parallel, rather than that they don't work without Zoltan). I'll have a go at writing the additional code and add zoltan tags. Aside from that, the only outstanding failures are the two medium tests with quadrature issues, where I admit I'm lacking inspiration.

jrper:
To update those following the sam issues at home: after further testing it appears that the current implementation of libsam will work for flredecomp with discontinuous elements. However my last commit  (8f860db) adds the zoltan tag to 5 tests which were failing without zoltan for nontrivial reasons.
In the short tests:
detectors_parallel_init belonging to @mlange05 now fails due to options checking and the fact that it specifies options to pass to zoltan (the test itself passes with sam, so perhaps the options are unnecessary??)
heat_2d_dg_local_assembly.xml belonging to @wence fails due to generating empty partitions with libsam. 
In the medium tests geopressure_convergence & geopressure_convergence_p1 belonging to @matt-piggott and viscous_fs_zhong_spatial belonging to @cianwilson  all appear to fail with libsam due to issues connected to python import statements and memory issues. Ideally it would be nice to clear this up, but I suspect it would taie more time than I have available to give it, unless @ggorman wants to take a look.

@tmbgreaves Is there any chance of getting a second buildbot queue for this with working zoltan to double-check the default behaviour is unchanged?

tmbgreaves:
Done - there are now two queues:

http://buildbot.ese.ic.ac.uk:8080/builders/sam_fix_nozoltan
http://buildbot.ese.ic.ac.uk:8080/builders/sam_fix_zoltan

Should be identical except that the nozoltan version has the zoltan headers and library removed to 'break' Zoltan.

stephankramer:
I've just been reminded of another thing I'd like to change about how we deal with libsam. If in the default ./configure of fluidity it fails to find zoltan I think it should end the configure with an error (perhaps suggesting to configure explicitly asking for sam instead), instead of automagically switching to sam as it does no. A lot of people come in for a surprise when they think they've successfully build only then find out at run time that they don't actually have all the standard functionality of fluidity.

If you don't mind I'm willing to make that change and add it to this branch. Or if you prefer, I can do it later in a separate branch.

stephankramer:
Is this good to go then? Does someone else need to look at it?

jrper:
I think I see the intent of your changes, but I'm afraid I'm not confident enough in my understanding of the black magic of autotools to confirm them to be correct. On the other hand, it has a green buildbot queue with no failures and 20 addition passes compared to the trunk nozoltan so it's obvious still doing something more correctly.

tmbgreaves:
I am happy with the autotools changes - a very neat job, thanks Stephan.

stephankramer:
This has now been merged

PR #55, opened 2015-05-18
Flredecomp file type fix

stevendargaville:
This branch fixes some problems with flredecomp to allow the use of arbritrary schema. Even if the schema has changed significantly compared to the Fluidity base schema, at the very least this change allows you to still decompose a mesh without using fldecomp.

This came out of the Zoltan/PETSc/fldecomp discussion.

### DIFF
--- a/tools/Flredecomp.F90
+++ b/tools/Flredecomp.F90
@@ -118,8 +118,8 @@ subroutine flredecomp(input_basename, input_basename_len, output_basename, outpu
      output_file_type=input_file_type
   end if
   
-  ewrite(2, "(a)") "Input base name: " // trim(input_base)
-  ewrite(2, "(a)") "Output base name: " // trim(output_base)
+  ewrite(2, "(a)") "Input base name: " // trim(input_base), " type " // trim(input_file_type)
+  ewrite(2, "(a)") "Output base name: " // trim(output_base), " type " // trim(input_file_type)
   ewrite(2, "(a,i0)") "Input number of processes: ", input_nprocs
   ewrite(2, "(a,i0)") "Target number of processes: ", target_nprocs
   ewrite(2, "(a,i0)") "Job number of processes: ", nprocs
@@ -138,8 +138,9 @@ subroutine flredecomp(input_basename, input_basename_len, output_basename, outpu
   end if
   
   ! Load the options tree
-  call load_options(trim(input_base) // trim(input_file_type))
-  if(.not. have_option("/simulation_name")) then
+  call load_options(trim(input_base) // trim(input_file_type))F
+  ! Ignore this error if we're not processing fluidity schema 
+  if(.not. have_option("/simulation_name") .and. trim(input_file_type) == "flml") then
     FLExit("Failed to find simulation name after loading options file")
   end if
 
@@ -170,6 +171,9 @@ subroutine flredecomp(input_basename, input_basename_len, output_basename, outpu
   
   ! Find out how many states there are
   nstates=option_count("/material_phase")
+  ! If we're not processing fluidity schema, then we may not have states
+  ! in the same fashion as fluidity
+  if (nstates == 0 .and. .not. trim(input_file_type) == "flml") nstates = 1
   allocate(state(1:nstates))
   do i = 1, nstates
      call nullify(state(i))
--- a/tools/Flredecomp.F90
+++ b/tools/Flredecomp.F90
@@ -140,7 +140,7 @@ subroutine flredecomp(input_basename, input_basename_len, output_basename, outpu
   ! Load the options tree
   call load_options(trim(input_base) // trim(input_file_type))
   ! Ignore this error if we're not processing fluidity schema 
-  if(.not. have_option("/simulation_name") .and. trim(input_file_type) == "flml") then
+  if(.not. have_option("/simulation_name") .and. trim(input_file_type) == ".flml") then
     FLExit("Failed to find simulation name after loading options file")
   end if
 
@@ -173,7 +173,7 @@ subroutine flredecomp(input_basename, input_basename_len, output_basename, outpu
   nstates=option_count("/material_phase")
   ! If we're not processing fluidity schema, then we may not have states
   ! in the same fashion as fluidity
-  if (nstates == 0 .and. .not. trim(input_file_type) == "flml") nstates = 1
+  if (nstates == 0 .and. .not. trim(input_file_type) == ".flml") nstates = 1
   allocate(state(1:nstates))
   do i = 1, nstates
      call nullify(state(i))
--- a/tools/Flredecomp.F90
+++ b/tools/Flredecomp.F90
@@ -138,7 +138,7 @@ subroutine flredecomp(input_basename, input_basename_len, output_basename, outpu
   end if
   
   ! Load the options tree
-  call load_options(trim(input_base) // trim(input_file_type))F
+  call load_options(trim(input_base) // trim(input_file_type))
   ! Ignore this error if we're not processing fluidity schema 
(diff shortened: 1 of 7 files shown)


stephankramer:
There seem to be various random changes in indentation in a lot of files. Could you fix these first please? It's kind of hard to read what has actually changed - also we shouldn't be unnecessarily touching lines of code.

jrper:
Apologies, the whitespace came from my slightly overeager use of meld. Hopefully fixed in this commit, to be squashed before merging (@stevendargaville ar you happy to do this?)

tools/Flredecomp.F90
```diff
@@ -118,8 +138,9 @@ end subroutine python_init
   end if
   
   ! Load the options tree
-  call load_options(trim(input_base) // ".flml")
-  if(.not. have_option("/simulation_name")) then
+  call load_options(trim(input_base) // trim(input_file_type))
+  ! Ignore this error if we're not processing fluidity schema 
+  if(.not. have_option("/simulation_name") .and. trim(input_file_type) == "flml") then
```
stephankramer: should that be ".flml"?

tools/Flredecomp.F90
```diff
@@ -147,6 +171,9 @@ end subroutine python_init
   
   ! Find out how many states there are
   nstates=option_count("/material_phase")
+  ! If we're not processing fluidity schema, then we may not have states
+  ! in the same fashion as fluidity
+  if (nstates == 0 .and. .not. trim(input_file_type) == "flml") nstates = 1
```
stephankramer: Again ".flml"?

stephankramer:
OK, great that's a lot more readable.

Just to get some discussion started, can I summarize that what this implements is the ability to run flredecomp on a non-flml schema input file, where the non-flml schema: 1) does not have /problem_type 2) does not have /material_phase 3) the extension is not ".flml" ? Anything else?

Just out of curiosity if it doesn't have /material_phase does that mean we're not flredecomping any fields, only meshes under /geometry/ ?

About adding the stat argument to check_options(): if all we're trying to do is circumvent the check for /problem_type I think I'd be happier for that check to be removed, we could always re-add it in more fluidity specific parts of the code. Otherwise I think we have to discuss a little more about the semantics of this stat argument: *) normally providing a stat argument means: don't ever error but give me back an error code *) it should have an intent, but what should it be; what happens if multiple check_options subs have a stat argument, etc. etc.

jrper:
My changes were aimed purely at making it file extension agnostic, but I believe @stevendargaville needs/wants to use flredecomp for a wider set of problems in which there is only a mesh. It's probably better if he confirms that though. Thanks for looking at this again, I believe the issues you pulled up were correct and have now been corrected.

From memory, the addition of a stat argument was proximately to silence the "problem_type" issue, but also as a half-hearted method to allow similar opt-outs for other options checks, since it wasn't at all obvious that the collection called was necessary for all possible codes which might want to use flredecomp. I'm happy to have this conversation if its useful, but I'm not sure which medium would include all interested parties, since there are several different groups piggybacking off the Fluidity code base, and I'm not sure there's a venue which includes them all.

stevendargaville:
So we definitely need flredecomp to work when there is only a mesh present. I think this is true of many of the codes that live on top of Fluidity, where the schemas/fields/problem_types may have changed so much that the only common thing is the mesh.

Basically we only require flredecomp to decompose the mesh at the start of our simulations. At present we're using fldecomp, but given the future of fldecomp was a little unclear for a while this seems like a sensible thing to enable in flredecomp. 

We won't specifically be using flredecomp to change the number of processors after checkpointing for example. I'm not sure if that is true of other codes, but the changes in this commit at least allow us (and anyone else dependent on fldecomp) to jump to flredecomp without any decrease in functionality.

stephankramer:
Ok, basically what I want to establish is: what is the minimum set of requirements on an input .xml for which we want to make flredecomp work? This should be well defined and not something like "well it doesn't chance the current behaviour and this chance makes it do what I want".

I'm still a little confused as to how you can not have /material_phase. All that flredecomp does is call populate state, call zoltan to redistribute, and then call checkpoint. If you call populate state without /material_phase it allocates 0 states. It might still loop over /geometry/mesh and populate your meshes, but then it has nothing to put these meshes in. What am I missing here?

stevendargaville:
So for us, the minimum input xml we would like to work with flredecomp is simply a /geometry/mesh so we can statically decompose a mesh.

We don't have any /material_phase in our schema (that's actually not always true, we run coupled fluids/radiation simulations that have the current fluidity schema included in our xmls, but that is not important), and because of this we have our own version of populate_state that gets called in our code. This is why we can't use the populate_state the flredecomp would call, as it is the trunk Fluidity version.

Does this help clarify?

stephankramer:
Great, thanks for the clarification. As usual however, this throws up new questions: if you are actually using a different populate_state, then I'm a bit confused why we are discussing a change in the check options system that only applies in populate_state.

Just to be clear, I don't mind changing what options checks are run with flredecomp - I'd probably would like to do it slightly differently - but it's probably better to clear up first what you want. In any case, any requested functionality should be testable within the fluidity master framework.

tmbgreaves:
Steven, James, could you update on whether you're still wanting to have this pulled into master please?

jrper:
It's really up to @stevendargaville, now that I'm only working with vanilla fluidity day to day. I guess there are wider questions about how to support people using femtools but not fluidity on which @aladam might also have views, but these probably shouldn't get sorted out on a pull request.

stevendargaville:
So I'd still be keen to have this merged, but I'm not sure I satisfied @stephankramer previously.  I would be happy to add a test which checks that flredecomp can decompose a mesh given a (vaguely) arbitrary schema (we could just use one of our generic schema). This is the only functionality I need from this merge, and I'm sure at some point it would be useful for other codes that sit on top of fluidity.

tmbgreaves:
I think @stephankramer is entirely offline this week but will wait on his comments when he returns.

stephankramer:
Ok, so let's continue this discussion then. I still don't understand how you can not have /material_phase at all and expect flredecomp to do something sensible. You say you have your own version of populate_state(), which presumably deals with this somehow - but then why do the proposed changes need to go into main fluidity?

stevendargaville:
So our use case is purely to decompose a mesh, which we normally use fldecomp for. We don't ever use flredecomp to "redecompose" a mesh and it's associated fields, as you rightly pointed out, our own version of populate state and lack of material_phase makes that outside the scope of main fluidity. 

The two things that motivated this change was the discussion over the future of fldecomp and the ability to decompose a mesh in parallel. If fldecomp goes away, we would completely lose the ability to decompose a mesh. The ability to decompose a mesh in parallel with the modified flredecomp would also be very handy for us. 

Does that answer your question @stephankramer?

stevendargaville:
I guess fundamentally the question might be, do you think it is valuable to have flredecomp behave like fldecomp in absence of anything but a mesh?

tmbgreaves:
Following up from today's dev meeting, the feeling was that despite the maintenance overhead of continuing to directly track metis interfacing (which we'll do via the PETSc internal build of metis) we will commit to maintaining fldecomp in the long term. Resulting from this was a decision to remove the notice in fldecomp that it is deprecated, and resulting from that was a suggestion that this pull request be closed. Steven, would you be happy with that outcome?

stevendargaville:
Sure that sounds fine, as long as fldecomp remains supported that fulfils the majority of the functionality I need. Thanks everyone.

jrper:
I think everyone's happy with the resolution, but can I suggest we keep the branch in collective memory, just in case the fldecomp support becomes too high at a point in time before everyone is using petsc/firedrake to manage meshes anyway?

tmbgreaves:
Sounds good to me. Closing now but remembering.

PR #63, opened 2015-07-22
Fix momentum subcycling weak bcs

stephankramer:
Fix weak bcs for momentum subcycling.
    
Previously the necessary terms associated with a weak Dirichlet bc for advection on the rhs (the ones containing the actual bc values) where left out entirely when using subcycling. These are now applied as a rhs term during the subcycling.
    
This potentially changes any run with DG subcycling for momentum with a weak Dirichlet inflow condition. Since the condition for the normal component is also imposed via the continuity equation, it may have done roughly the right thing for cases where advection is not dominant at the open boundary. However local artefacts would still have been likely - apparent through an increased pressure along the open boundary, and a noisy velocity.

Switch on subcycling in friction-slope-swe test (this test would fail previously with subcycling).

Remove acceleration logical in momentum_dg that was never used, i.e. always assumed true, to simplify the code a little.

### DIFF
--- a/assemble/Momentum_DG.F90
+++ b/assemble/Momentum_DG.F90
@@ -3054,7 +3054,7 @@ contains
         ! filter wiggles from u
         do d =1, mesh_dim(u)
         u_cpt = extract_scalar_field_from_vector_field(u_sub,d)
-        !call limit_vb(state,u_cpt)
+        call limit_vb(state,u_cpt)
         end do
 
       end if
--- a/assemble/Momentum_DG.F90
+++ b/assemble/Momentum_DG.F90
@@ -111,9 +111,6 @@ module momentum_DG
   ! Parameters for interior penalty method
   real :: Interior_Penalty_Parameter, edge_length_power, h0
 
-  ! Flag indicating whether equations are being solved in acceleration form.
-  logical :: acceleration
-
   ! Flag indicating whether to include pressure bcs (not for cv pressure)
   logical :: l_include_pressure_bcs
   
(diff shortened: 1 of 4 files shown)


assemble/Momentum_DG.F90
```diff
@@ -2319,6 +2297,13 @@ subroutine construct_momentum_interface_dg(ele, face, face_2, ni, &
                subcycle_m_tensor_addto(dim, dim, u_face_l, start:finish) = &
                     &subcycle_m_tensor_addto(dim, dim, u_face_l, start:finish)&
                     &+nnAdvection_in
+            else
```
cianwilson: Where the main magic is happening (aside from all the acceleration changes)?

assemble/Momentum_DG.F90
```diff
@@ -3085,9 +3060,11 @@ subroutine subcycle_momentum_dg(u, mom_rhs, subcycle_m, inverse_mass, state)
       end if
 
  
-      ! du = advection * u
+      ! du = advection * u - f_adv
       call mult(delta_u, subcycle_m, u_sub)
-      ! M*du/dt = M*du/dt - advection * u
+      ! -f_adv for bc terms
```
cianwilson: Another bit of the magic?

cianwilson:
Added a couple of line comments just to make sure I've seen the relevant bits wrt subcycling bcs (i.e. aside from all the acceleration form changes).

Otherwise I see a green buildbot [here](http://buildbot-ocean.ese.ic.ac.uk:8080/builders/fix-momentum-subcycling-weak-bcs) so it looks good to me.

stephankramer:
Thanks for having a look. That's indeed where the magic happens, other than it actually being added into a new subcycle_rhs field which is allocated and deallocated at the appropriate points in momentum_equation.

PR #65, opened 2015-08-05
Move matsetnullspace

stephankramer:
This merge fixes some issues introduced by switching from KSPSetNullSpace to MatSetNullSpace. It adds petsc 3.6 support, and it sets the nullspace on mat instead of pmat (consistent with petsc>=3.6.1)

When using MatSetNullSpace with petsc < 3.6 it should be done before KSPSetOperators is called, as this internally calls KSPSetNullSpace at that point to set the nullspace that gets removed during the krylov iteration. For petsc>=3.6 this doesn't matter as nullspaces are no longer attached to ksps, and the nullspace is simply picked up from the matrix during the iteration.

As upon our request in petsc>=3.6.1 the nullspace is no longer picked up from pmat, but from mat instead. This is to avoid having to set a nullspace for cases where mat does, but pmat doesn't have a nullspace. This is the case for instance in  Stokes_3d_busse_1a_p2p1, where the Schur complement of the Stokes system has a null space, but we approximate this with a "scaled pressure mass matrix" that does not have a nullspace. Setting the nullspace on the outer solver, would previously require setting it on pmat, which would automatically mean it would be applied in the pcksp preconditioner solve as well, leading to spurious results.

Thus in Fluidity we now call MatSetNullSpace on mat instead of pmat to conform with petsc>=3.6.1. For petsc<3.6 we have introduced an extra KSPSetNullSpace to ensure the right nullspace is picked up. For petsc 3.6.0 (the first 3.6 release before 3.6.1) this means we don't support the case where mat/=pmat and the mat and pmat have different nullspace (an error is given). In other words petsc 3.6.0 should work for most cases, but some tests will fail and is therefore not fully supported.


stephankramer:
I just saw two random unrelated mesh changes have crept in:
 tests/optimality/src/mesh_A.msh 
 tests/shallow_water_optimisation_check_gradient_2d/src/mesh_A.msh 
will revert those when I rebase this.

cianwilson:
Looks good to me (this has already been reviewed once offline).

stephankramer:
Merged

PR #66, opened 2015-08-24
Fluidity zoltan lb approach

drhodrid:
Fluidity master currently enforces the load balancing approach "REPARTITION" on all adapt iterations, excluding the final adapt iteration, where we use the "PARTITION" approach. As a reminder, REPARTITION tries to stay close to the current partition/distribution, whereas PARTITION partitions from scratch. I guess the previous logic was that repartition should be more efficient, as it should minimize data communication. 

However, I have found, for several (if not all) simulations that I've been running, that PARTITION turns out to be significantly more efficient. The desired quality mesh is reached with fewer adapt iterations and this turns out to be significantly faster. 

With this in mind, in this branch, I have changed Fluidity's default load balancing approach from repartition to partition, for all adapt iterations. However, the load_balancing approach is now an optional option inside the .flml, so that users can change this, if desired. Note that I have not changed the default behaviour (or added an option) for the final adapt iteration load balancing approach - this, in my opinion, should always be "PARTITION". 

An added bonus of PARTITION over REPARTITION is that I am seeing less "empty partition" issues with zoltan... which is a big plus. 

Could somebody set up a buildbot queue for this branch?


tmbgreaves:
http://buildbot.ese.ic.ac.uk:8080/builders/fluidity-zoltan-lb-approach

stephankramer:
When you say that it's significantly faster, how did you measure that? Also is this in combination with a higher number of adapt iterations than the default (i.e. is it being faster due to there being less iterations)?

drhodrid:
Hi Steph,

My measure of speed is the total simulation time. For the examples I've run, the larger the number of cores, the larger the speedup of Partition relative to Repartition. As an example, for a 3D Adaptive Stokes Case on 64 cores, with 2 materials and a complex rheology, I see a speedup of around 5%. On 128 cores, I see a speedup of around 18-20% for the same case. These were both run on the same (empty) system. The speedup seems related to 2 things (note that these aren't always true):

1. Partition typically takes less adapt iterations to return the desired quality mesh than Repartition (not always, but generally).
2. If all adapt iterations are used (and we still have not reached the desired mesh quality), the mesh quality from the Partition setup is almost always superior to the Repartition setup, so the solvers are marginally quicker. 

As you noted, I am running these with more adapt iterations than the default: 3 + 7, versus 3. For me, having a good quality mesh post adapt is essential: if I have a bad element, my simulation is worthless, so I'd rather be safe. 

One other point: I've had a few emails since the commit getting excited about this merge getting rid of Fluidity's "Empty partitions" problem. I wanted to emphasise that it helps, but it is NOT the full solution - Zoltan does not guarantee empty partitions and sometimes things still break down in fluidity because of this. From my experience, setting the partitioner and final partitioner to scotch and setting a tight load imbalance tolerance of 1.02 helps with this issue (but again, it doesn't always go away). 

Note that buildbot is currently red, so I'll look into this today.

tmbgreaves:
Superseded by #94

PR #70, opened 2015-10-23
Reimplement radial extrusion within the original extrusion code

cianwilson:
This pull request removes the radial extrusion code and reimplements all its tested functionality in the original code path.  This significantly tidies up extrusion code.  The radial code path had originally been copied and pasted anyway and then, after initial small modifications to make it work radially, it slowly diverged making adding new functionality difficult as it needed to be implemented in two places (or have ugly comments added to the schema saying it didn't work with a particular version).

This has the added benefit of making options that weren't available in radial but were available in planar available to both.  This includes the only_varies_in_depth (renamed from only_varies_in_z) option and the surface_height option.  Additionally sigma layering was not properly implemented in radial (for large variations in bathymetry) but should now work.  Finally, radial extrusion made 3d assumptions.  These have now been removed and a 2d test has been modified to check this.

Green buildbot branch [here](http://buildbot-ocean.ese.ic.ac.uk:8080/waterfall?show=merge_spherical_planar_extrusion).


stephankramer:
Looks great. Couldn't find a thing to moan about...

cianwilson:
Thanks @stephankramer.  Rebased and merged.

PR #71, opened 2015-10-23
Allow extrusion upwards by setting negative depths

cianwilson:
Previously extrusion was only able to extrude down to a particular depth.  This pull request allows negative depths to be set that then extrude upwards.  The sizing function should still be positive.

Two test have been modified to extrude upward (one radially and one planar) to test this functionality. 

Green buildbot branch [here](http://buildbot-ocean.ese.ic.ac.uk:8080/builders/upwards_extrusion).

This pull request should not be merged until #70 has been merged as this is a branch off the merge_spherical_planar_extrusion branch.


stephankramer:
Great. Surprisingly small diff (when comparing with merge_spherical_planar_extrusion). Tested and all - good to go!

cianwilson:
Thanks @stephankramer. Rebased and merged.

PR #72, opened 2015-10-23
Modify incorrect options check.

cianwilson:
You should be able to exclude meshes from adaptivity that are extruded from.  This pull request modifies an options check that previously prevented this.

This allows an extruded mesh to be adapted in an unstructured way (i.e. when vertically structured adaptivity is not turned on), destroying the effects of the extrusion but without crashing because there's another lower dimensional mesh hanging around.


stephankramer:
Agreed. Again, here the underlying issue is that there's apparently no tests for extrusion + normal adaptivity. Have opened an issue, #77.  This PR can of course be merged straight away.

cianwilson:
Rebased and merged.  Thanks Stephan!

PR #73, opened 2015-10-29
Remove pcfieldsplitgetsubksp FORTRAN wrapper 

jrper:
… and update the code to match the existing PETSc one.


stephankramer:
Looks great and a good way to break with older petsc versions.

jrper:
Green on buildbot (with PETSc 3.4, so I think this really does only break PETSc versions from pre 2012) merged and closed.

PR #74, opened 2015-11-04
Change exit state of testharness following completed test runs

tmbgreaves:
Prior to this merge, testharness can complete test runs but still exit with error if test it runs returns failures. This can be confusing as it suggests that testharness itself has experienced an error - a state which occurs if (for example) an invalid xml file is specified in the input. This commit removes the errant behaviour, and puts the onus on output parsers to detect whether or not tests are passing.

### DIFF
--- a/tools/testharness.py
+++ b/tools/testharness.py
@@ -331,10 +331,6 @@ class TestHarness:
             print "Passes:   %d" % self.passcount
             print "Failures: %d" % self.failcount
             print "Warnings: %d" % self.warncount
-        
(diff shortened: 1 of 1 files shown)


jrper:
That looks like the only necessary change to me, and the new behaviour makes commands like
make test && make mediumtest behave more understandably. 

I suppose it might be worthwhile to make a command line option to return the fail count as the exit code for people depending on the existing behaviour, or for easy script parsing, but I could wrap that into the branch I've been working on to do xml outputting.

PR #75, opened 2015-11-13
Changes to project to continuous

stevendargaville:
This change passes all tests. Any thoughts?


stephankramer:
Agreed. This brings up a different issue: there's no tests for project_to_continuous - so this tool is likely to bitrot or could legitimately be removed. Fine to merge this now, but I've opened up an issue (#76) about the lack of tests.

Issue #50, opened 2015-05-05
check_options routines with brackets ignored in code auto generation


jrper:
Apologies if this is documented somewhere and I just couldn't find it, but while investigating changing flredecomp to work better with arbitrary file names, I've discovered that due to the regexps in the make_check_options script, module subroutines which end with parentheses are currently excluded from the auto generated check_options.F90 file. Is this behaviour deliberate, or a bug? If the latter, then a couple of modules also need patching:
1) to make the routines public
2) to let complete_field_path know about aliased fields.

jrper:
I think I've pulled out the relevant code into a branch (939813403385869c6dcba2076328b8cb149434ec), but I guess a pull request should wait on confirmation of the correct behaviour.

stephankramer:
Good catch, that really doesn't look like it should happen indeed (these _check_options() routines being ignored). I guess fixing this you run the chance of breaking some tests that are currently failing the assumptions of these tests (rightly or wrongly). Is that what your point about complete_field_path is about? Could you expand a bit more on that?

jrper:
It's probably easiest to see in a code diff, but the current version of complete_field_path fails with an error when applied to aliased fields (it's just not a listed field type in a case select). This doesn't seem to matter anywhere that's currently tested, but when the options check in assemble/Interpolation_manager is actually being run it applies the routine to all known fields to check super meshing options, and the way it's done means any tests involving aliased fields fail. Patching that as in the branch above gives no reported failures on short tests run locally on my workstation, but I realise that's by no means all possible tests.

cianwilson:
Do aliased fields have independent option paths? My vague recollection is that they get the option_path of the original field and we tell if they're aliased using an aliased flag in the field type definition. This may be the reason for the apparent oversight in complete_field_path.

jrper:
Yes, that seems to be the case in general. It looks like the Interpolation options check code is the only place where that routine is applied to the raw options tree rather than a post-processed option path. I'm not sure whether that means the correct fix is to rewrite the logic in the routine to be a bit less optimistic.

PR #80, opened 2015-11-24
Changes to allow all check_options tests to run.

jrper:
This is probably worth a pull request. It's the code written to deal with issue #50, where check_options routines are only auto detected when written in a specific pattern, whereas other patterns are both valid fortran, and used in the code base. It'll need a buildbot queue.


tmbgreaves:
http://buildbot.ese.ic.ac.uk:8080/builders/check_options_cleanup

PR #81, opened 2015-11-24
Add exit status and xml reporting options to test harness

jrper:
These two changes introduce extra command line options to test harness

The first (requiring the junit_xml python module) produces junit formatted xml reports of the test results.
The second provides an option to set the exit status to the failure count, allowing people to revert to the old behaviour in any automated scripts.


tmbgreaves:
This all looks good and useful to me; thanks James!

tmbgreaves:
Passing on buildbot; will check fails also report correctly once dockerhub stops 503ing me.

tmbgreaves:
Checked fails also report correctly; they do.

PR #83, opened 2015-11-29
Strip l2 halo

stephankramer:
This reimplements an idea present in one of @sparkinson 's branches, which is to strip out the L2 halo before going into adapt. This results in fewer locked elements during parallel adaptivity - thus improving issues with poor mesh-quality parallel halo regions or the need for an excessive amount of parallel iterations. This implementation does not use libsam's halo stripping code but introduces new femtools routines to do this.

Other changes:
- in 2d, the amount of locking has been reduced even further. Even after stripping the l2 halo, the libmba2d wrapper would still lock elements adjacent to partially owned elements - presumably to avoid that changes in those adjacent elements would affect the partially owned elements - this seems unnecessary however since we already explicitly tell libmba2d to lock these. In effect this means that now we only have single strip of elements that is locked (any elements with nodes owned by different processes) - whereas before at least 5 strips where locked (the mixed-owner elements and two rows on either side).
- in 3d, the convergence criterion for parallel adaptivity iterations has been improved. Previously the convergence criterion would use the Lipnikov functional to determine whether the quality of the mesh was sufficient everywhere. The 3d adaptivity library however uses the "Pain" functional so that a mismatch between the two could lead to the iteration not converging at a point where the adaptivity library itself thinks it has done enough.
- in elements that are locked change the metric that is fed to the adaptivity library to pretend all locked elements are already of sufficient quality. This is to avoid adaptivity library giving up altogether (as observed for at least libma2d), as it thinks the quality criteria cannot be met. This only affects the metric that is fed to the adaptivity process within each local domain separately, the metric of the locked regions is kept intact in the global metric - so that the mesh will be improved when these regions are unlocked again in the subsequent iterations
- add tests that test coarsening and refinement in 2d and 3d parallel adaptivity without leaving halos regions that are not sufficiently coarsened/refined.


jrper:
I can't spot anything incorrect with this. There are a couple of instances of repeated code where it looks like it would be nice to have an element_halo_level function. I also wonder whether the default numbering with halo nodes at the end could be leveraged more generally, but robust is probably better than optimised given the labyrinth of code paths we currently support.

stephankramer:
Thanks @jrper for reviewing this. As discussed, I had a look at repeated element-node ownership logic. There are indeed a couple of repeated loops over element nodes with some sort of logic about node ownership applied. Unfortunately that logic is slightly different in all of them, so I don't think you can summarize them in a single element halo level. For instance elements that are stripped, are elements of which all nodes are not owned, this does not necessarily mean they are halo 2 - as there are corner cases with elements that have only halo 1 nodes (in particular on corners of subdomains). Elements that are locked are elements that have at least one non-owned node, etc. Note that there is already a separate element halo level derived from the %element_halos, which is strictly about mutually exclusive ownership of elements and the question who is responsible for which bit of the DG assembly. Wrt leveraging trailing-receive numbering, we discussed this offline and agreed that in this case robustness is to be preferred over premature optimisation.

PR #88, opened 2016-02-10
Fix parallel extrude and adapt

stephankramer:
Fix for general (non 2plus 1) adaptivity after initial mesh extrusion.

Main fix is to tell zoltan which mesh it needs to decompose: not the horizontal mesh, but the adapted, (initially) extruded mesh. This means we will leave the input horizontal mesh dangling, and keep it in the old decomposition.

Added a test for this setup including a checkpoint from an adapted mesh.


jrper:
All looks good to me. Assuming the green buildbot queue is from post @drhodrid's merge, I'm happy for this to go in.

tmbgreaves:

tmbgreaves:

Issue #87, opened 2016-02-05
Ordering of module use statements


jrper:
As I believe is common knowledge, the Intel Fortran compilers still appear to be very sensitive to the order of use statements in modules, with compiler errors liable to appear if these get too far away from the INTEL preferred order (if both a module and another which uses it are to be included, the parent must be included before the children).

Given that I'm currently running into problems with this while trying to use a branch on the imperial cluster, I'm proposing to fix this for as many modules as I need to, while also including a python script to detect this behaviour if this starts to be a problem for others. Since we also have external users running on intel, and others may have views on what they'd like to have happen I'm raising this as an issue now first.

dham:
I don't have anything to offer other than sympathy. This has been an ifort bug for the best part of a decade now!

jrper:
For anyone keep ing track at home branch 0868fdd27e899ef6ca1fbecc37427c69949b4a2b now has the work in it. Since my script accidentally ended up more powerful than I intended, I've applied it to everything. I'm now working on a test to go with this, then I'll open a pull request. If anyone has any idea how to deal with timing compiles under intel, I'd be interested whether this version is also faster, or  only avoids internal compiler errors.

jrper:
Closed via #96.

PR #89, opened 2016-02-11
Use statement cleanup

jrper:
As described in #87, this changeset reorders use statements in all modules to be in the Intel fortran compiler preferred order. This in turn gives slightly more overhead before running into internal compiler errors when additional code is added. A warn test has also been added to scan the ordering of the files as additional work is added.

### DIFF
--- a/assemble/Adapt_State.F90
+++ b/assemble/Adapt_State.F90
@@ -28,56 +28,53 @@
 #include "fdebug.h"
 
 module adapt_state_module
-! these 5 need to be on top and in this order, so as not to confuse silly old intel compiler
+  use global_parameters, only : OPTION_PATH_LEN, periodic_boundary_option_path, adaptivity_mesh_name, domain_bbox, topology_mesh_name
   use quadrature
   use elements
+  use mpi_interfaces
+  use parallel_tools
+  use data_structures
   use sparse_tools
+  use eventcounter
+  use intersection_finder_module
   use fields
   use state_module
-!
+  use halos
+  use field_options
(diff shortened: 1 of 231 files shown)


jrper:
Yes, I came to that realisation shortly after I managed to get the current version working. Do you consider that the maintainability benefits demand a rewrite?

wence-:
Nah, probably not.

assemble/Adapt_State.F90
```diff
@@ -28,56 +28,53 @@
...
-  use data_structures
-  use detector_data_types
-  use detector_parallel
-  use diagnostic_variables
-  use intersection_finder_module
-  use diagnostic_variables
-  use diagnostic_fields_wrapper_new, only : calculate_diagnostic_variables_new => calculate_diagnostic_variables
-  use pickers
   use write_gmsh
 #ifdef HAVE_ZOLTAN
```
stephankramer: Yeah, so I was wondering about this: I presume you currently ignore `#ifdef` lines? So in this specific case the result of moving `mpi_interfaces` out of the `#ifdef HAVE_ZOLTAN` block is harmless - but how does this work generally?

jrper: Currently preprocessor protection is indeed ignored when the script calculates the necessary ordering. It's up to a human to put it in as necessary, and I missed it for `mpi_interfaces` here. That should be ok providing we never set up circular dependencies in the unpreprocessed module files, which I can't see a reason to do.

jrper:
Included as part of #96.

PR #90, opened 2016-02-15
Various fixes for the unifiedmesh tool

cianwilson:
Seg faults from unifiedmesh were causing failures of the explicit-hyperc-shear-adapt test case.  This pull request tries to fix them.

Firstly, iso_c_bindings were not being used so that's been fixed.

Secondly, unifiedmesh writes a gmsh file but it uses a mesh without faces.  Write_GMSH assumes you have faces so it can find sloc but it doesn't actually need to know sloc because there are no faces.  The fix here is just to add logic about the number of faces wherever sloc is used.


stephankramer:
Beatiful. I'll add a buildbot branch - otherwise good to go.

cianwilson:
Thanks Stephan.  Buildbot went green so rebased and merged.

PR #93, opened 2016-02-16
Load imbalance tol

drhodrid:
This commit alters the load imbalance tolerances that Fluidity passes to zoltan. The default load imbalance tolerance is set to 1.05 (modified from 1.5) and the final adapt iteration tolerance is set to 1.02 (from 1.075). I have found (over the past 3 months) that these changes lead to a better balance in the final adapted mesh.

Green buildbot queue here:
http://buildbot-ocean.ese.ic.ac.uk:8080/builders/load-imbalance-tol


mlange05:
I have needed to change these in the past for highly parallel runs. Would it make sense to expose the tolerances as user options in the .flml with the proposed values as defaults?.

drhodrid:
Hi Michael,

You can already set the  load imbalance tolerance inside the .flml (under zoltan options I think). However, you cannot set the tolerance for the final iteration. I don’t really see a motivation for doing so, although I’m happy to add the option if others feel it’d be useful.

R


On 17 Feb 2016, at 5:32 PM, Michael Lange <notifications@github.com<mailto:notifications@github.com>> wrote:


I have needed to change these in the past for highly parallel runs. Would it make sense to expose the tolerances as user options in the .flml with the proposed values as defaults?.

drhodrid:
Hi all - any objection to this going in?

jrper:
Hi @drhodrid. If you need a reply form everyone this might take a long time, but I'm happy for it to go in if buildbot is green,

stephankramer:
I'm fine with merging this - but I do have some remarks: I presume the loose tolerance for the initial iterations is to give it a bit more flexibility to move the locked halo regions to the interior of local domains. I expect that tightening this tolerance may make problems with locked halo regions not being properly adapted a lot worse (than it already is).

You say "these changes lead to a better balance in the final adapted mesh"; How did you measure that?

tmbgreaves:

PR #94, opened 2016-02-16
Zoltan lb approach

drhodrid:
This updates a previous pull request, which is here:

https://github.com/FluidityProject/fluidity/pull/66 

To a new branch (sans camel case), which has also passed through buildbot:

http://buildbot-ocean.ese.ic.ac.uk:8080/builders/zoltan-lb-approach

As a reminder - Fluidity master currently enforces the load balancing approach "REPARTITION" on all adapt iterations, excluding the final adapt iteration, where we use the "PARTITION" approach. REPARTITION tries to stay close to the current partition/distribution, whereas PARTITION partitions from scratch. I guess the previous logic was that repartition should be more efficient, as it should minimize data communication.

However, I have found, for several (if not all) simulations that I've been running, that PARTITION turns out to be significantly more efficient. The desired quality mesh is reached with fewer adapt iterations and this turns out to be significantly faster.

With this in mind, in this branch, I have changed Fluidity's default load balancing approach from repartition to partition, for all adapt iterations. However, the load_balancing approach is now an optional option inside the .flml, so that users can change this, if desired. Note that I have not changed the default behaviour (or added an option) for the final adapt iteration load balancing approach - this, in my opinion, should always be "PARTITION".

An added bonus of PARTITION over REPARTITION is that I am seeing less "empty partition" issues with zoltan... which is a big plus.


drhodrid:
Could somebody take a look at this pull request please?

stephankramer:
My comments on this are gonna be related to #93: the rationale behind the previous settings (whether they work in practice or not) is that in the initial redistributions, the thing we care about most is moving locked halo regions to the inside of local domains without taking to much attention the load balance (adaptivity itself is usually fairly cheap) before the final adapt iteration. This could be achieved by setting a very loose load balance tolerance and using a cheap diffusive partitioner (which merely exchanges stuff with its neighbours). If you however used this mostly in conjunction with the tighened tolerances in #93 than the diffusive partitioner is going to struggle as it will not be able to both avoid edge cuts in the previously locked halos and adhere to the strict balance so it will arrive at some compromise which may not give the desired results (for instance some locked regions remaining locked) so that PARTION may seem to give better results.

Note that in the default fluidity setups we don't iterate the adapt iterations until a certain quality is reached. Talking about it being more efficient (in terms of number of adapt iterations) only really applies when you do. It also completely ignores the fact that the load balancing communication itself of  REPARTITION should be more efficient than PARTITION - so the cost *per* iteration could be less (but this can be counteracted by a worse load balance).

tmbgreaves:

PR #98, opened 2016-02-29
Replace Scientific.IO.NetCDF with scipy.io.netcdf

tmbgreaves:
python-scientific is now unsupported; it doesn't work with numpy >= 1.9 and there are no plans to fix this, hence it's being dropped from most distros as numpy upgrades to 1.9 and beyond. We only use this for Scientific.IO.NetCDF which appears to be able to be replaced like-for-like with scipy.io.netcdf.

This commit makes that change.

### DIFF
--- a/python/fluidity/diagnostics/era.py
+++ b/python/fluidity/diagnostics/era.py
@@ -26,9 +26,9 @@ import unittest
 import fluidity.diagnostics.debug as debug
 
 try:
-  import Scientific.IO.NetCDF as netcdf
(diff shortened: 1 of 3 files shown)


tmbgreaves:
Buildbot running at: http://buildbot.ese.ic.ac.uk:8080/builders/remove-scientific-python

tmbgreaves:
Green buildbot:

http://buildbot.ese.ic.ac.uk:8080/builders/remove-scientific-python/builds/0

PR #99, opened 2016-03-01
configure level test for intel advance bug.

jrper:
This commit adds an autoconf test to look for the intel advancing io bug. If found, it sets a preprocessor symbol such that the workaround code gets called, otherwise it is ignored. I _think_ this is the best way of dealing with the whole problem, given multiple compiler versions and Intel's habit of reintroducing old bugs into the latest versions of ifort.

### DIFF
--- a/configure
+++ b/configure
@@ -637,6 +637,7 @@ FLLINKER
 LINKER
 FORTRAN_REAL_8
 PROFILING_FLAG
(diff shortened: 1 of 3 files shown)


tmbgreaves:
Happy to sign off on this one - thank you James.

PR #100, opened 2016-03-02
Changing to scipy lower case namespace for NetCDF

tmbgreaves:
As James presciently pointed out, this one has come back to bite us: turns out that the scipy shipped with CentOS 6 doesn't support the upper case naming convention, hence changing to a lower case plus underscores version.

### DIFF
--- a/python/fluidity/diagnostics/era.py
+++ b/python/fluidity/diagnostics/era.py
@@ -25,10 +25,21 @@ import unittest
 
 import fluidity.diagnostics.debug as debug
 
-try:
-  import scipy.io.netcdf as netcdf
-except:
-  debug.deprint("Warning: Failed to import scipy.io.netcdf module")
+from scipy.version import version as SciPyVersion
+
+if SciPyVersion < '0.9.0':
+  try:
+    import Scientific.IO.NetCDF as netcdf
+  except:
+    debug.deprint("Warning: Your SciPy version is too old (<0.9.0) and \nthe Scientific.IO.NetCDF module failed to import")
+else:
+  try:
+    import scipy.io.netcdf as netcdf
+  except:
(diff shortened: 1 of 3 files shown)


tmbgreaves:
Buildbot running at:

http://buildbot.ese.ic.ac.uk:8080/builders/change_scipy_naming
http://buildbot.ese.ic.ac.uk:8080/builders/change_scipy_naming-centos6

jrper:
I'm happy for this to go in if it indeed deals with the problem of cents using a version of scipy from 2009. The gcc build appears to have run fine, but the cents build seems to be running into an unrelated problem.

tmbgreaves:
Hum, yes, there's something very odd going on with petsc_readnsolve which I've seen on a few queues now.

tmbgreaves:
This is turning into a can of worms - updating createDimension and createVariable to create_dimension and create_variable respectively, it moves onto further problems. I'm getting the feeling that a workaround may be in order here.

tmbgreaves:
Workaround added and branch rewritten to squash onto the previous attempt; re-running buildbot. This fixes for me running by hand so curious to see if it works on buildbot too.

tmbgreaves:
Green buildbot for Trusty and whilst the centos6 buildbot is having problems elsewhere I've run this by hand and with the latest fixes it passes. James, are you happy with my updated workaround-fix?

tmbgreaves:
Thanks James. (1) fixed and (2) attempting to offer a better error message. I think some people will be stuck with early scipy so it's potentially not very practical for them to upgrade this, and much easier for them to just install a Scientific package.

jrper:
Ok, I'm happy. Feel free to merge if you're satisfied with the centos behaviour.

tmbgreaves:
Thanks James!

PR #102, opened 2016-03-03
Update nonlinear fields

drhodrid:
This commit updates the way in which diagnostic fields are calculated inside momentum diagnostics. Rather than simply updating a field (e.g. viscosity) we now also update all of its dependencies (e.g. strain rate), thus keeping everything in sync and doing a better job of handling non-linearities. 

To maintain efficiency (i.e. to ensure that fields, which are dependencies for many other fields, are only updated on one occasion), we track the dependencies through a calculated mask, which requires some copying back and forward between two arrays of states (one for all states the other for the phase). This has been tested on a number of Stokes cases that involve viscosity / material viscosity / strain rates / different creep mechanisms. It seems to work beautifully. 

Green buildbot branch here:
http://buildbot-ocean.ese.ic.ac.uk:8080/builders/update-nonlinear-fields

And a big thank you to Cian for all of the clever bits!


jrper:
Out of interest, how much of this depends on magic naming conventions? Given Fluidity's mandate to do everything it's conceivable somebody might want to do visco-elastic non-Newtonian fluids one day.

drhodrid:
The reason for doing this was to get rid of (some) magically named fields. For example, we previously had StrainRateSecondInvariant as a magically named field, and this was updated prior to the viscosity inside momentum diagnostics. This is now handled automatically as a dependency.  

Provided that the field of interest (in our case Viscosity) is updated inside momentum diagnostics, its dependencies should now be handled correctly.

stephankramer:
Could you maybe add some more comments? I have no idea what `calculate_diagnostic_variable_dep` or `dep_states_mask` mean. What does `update_calculated_state` do versus `update_calculated_submaterials`?

stephankramer:
Thanks! I was going to complain about some hideous names like 'dep_states_mask', but I see that's copied from elsewhere - so maybe it's better to keep it consistent (horrid as it is).

One last suggestion: would it not be better to move both versions of reference_fields() into the state_module. You could actually just integrate it with the insert interface, as that's what it does really - insert all the fields of one (or multiple) state(s) into another.

cianwilson:
Can we add some sort of test to this?  Perhaps a steady Stokes problem, the viscosity for which is  a python diagnostic field that depends on something else being calculated that isn't calculated correctly until the first timestep?  Ugly I know.

drhodrid:
Hi Cian,

I have already included a test for this (Stokes slab detachment). This is a 1 timestep (20 non linear iteration) multi-material case, where the viscosity of one material depends on the strain rate second invariant. 

Rhod

cianwilson:
Oh, sorry Rhodri.  For some reason the pull request says no files in tests were changed.  Where can I find it and does it fail if someone breaks this functionality?

drhodrid:
Hey handsome - it was a preexisting medium test (in the tests directors). Formerly, we had hard coded the (magically named field) StrainRateSecondInvariant to be updated in momentum diagnostics. We removed that hard coding and this dependency is now calculated automatically and updated accordingly. And yes - this test fails if that dependency is not dealt with correctly.

cianwilson:
OK, so long as @stephankramer's happy, it looks good to me.  Thanks to whoever kicked the buildbot branch after my changes.

stephankramer:
I'm happy. Good to know this is tested indeed!

stephankramer:
Merged into master.

PR #104, opened 2016-03-07
Fix fladapt on CX1.

jrper:
It appears Fladapt.F90 is causing an internal compiler error on CX1. This commit fixes that file for intel.


tmbgreaves:
Thanks James! Looks good; running through buildbot now with cx1-binmod and cx2-binmod building off this branch, and a trusty builder at http://buildbot.ese.ic.ac.uk:8080/builders/fladapt-fix-cx1

tmbgreaves:
All good. Thanks James!

PR #105, opened 2016-03-10
Bugfixes: openmp build.

jrper:
Some additional changes to use statements required when compiling with --enable-openmp.


tmbgreaves:
All good, thanks James!

Issue #101, opened 2016-03-02
channel-flow-dg test : The hypre euclid preconditioner


jrper:
While looking through a buildbot log file on one of my branches, I noticed that this test is raising a lot of PETSC errors due to calling the euclid (i.e. LU) preconditioned from hypre, which appears to no longer exist in petsc 3.6. The test still passes, so it's obviously using _something_ to solve the linear system, but it would probably be nice if this were made explicit. The test is antediluvian and owned by @dham, so I'm not sure who to ask to look at this.

jrper:
For the record this all goes away if you switch the hypre preconditioned to boomerAMG.

tmbgreaves:
Resolved by pull request #106

PR #106, opened 2016-03-17
Change tests attempting to use the Euclid preconditioner

jrper:
As I reported in #101, It appears that hypre - euclid isn't a preconditioner option available in recent Petsc - at least as compiled by us. This change set switches the tests that use it to cg + boomerAMG / gmres+sor.

@tmbgreaves: Is there a way of just testing the test changes under the buildbot environment? It seems like overkill to build a whole branch for just this?

### DIFF
--- a/tests/channel-flow-dg/channel_viscous.flml
+++ b/tests/channel-flow-dg/channel_viscous.flml
@@ -140,7 +140,7 @@
         <solver>
           <iterative_method name="cg"/>
           <preconditioner name="hypre">
-            <hypre_type name="euclid"/>
+            <hypre_type name="boomeramg"/>
           </preconditioner>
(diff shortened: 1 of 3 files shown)


tmbgreaves:
I think policy is that we don't necessarily buildbot things that just touch tests, so as long as you're happy that this passes in trusty with our latest package set I'm happy for this to go straight in.

jrper:
Hmm, I've tested Trusty & petsc 3.6, but orthogonally rather than together. I'm fairly confident it should work, but I can't actually guarantee it.

tmbgreaves:
Given there are a good range of other possible combinations to test once in let's merge. At the worst we have three tests which could break for easily identified reasons.

PR #108, opened 2016-04-14
Bc pivot scaling rebased

stephankramer:
This PR is at the moment just for discussion with "team Stokes" to see what to do with this branch.

This branch brings in the following changes:
- a reordering of the DOFs for cg momentum to u1,v1,w1,u2,v2,w2,... instead of u1,u2,...v1,v2,...w1,w2,... This is to accomdate using GAMG on the velocity solve with a blocksize set (useful for Stokes problems). We keep the latter ordering for momenum dg, as that allows us to use a BAIJ matrix with BAIJ blocksize equal to nloc (DOFs per element) - which seems to give good performance with SOR preconditioning (as it does a block SOR)
- lifting of (strong) dirichlet bcs for momentum, getting rid of evil big springs
- the pivot-scaling refers to not using 1.0 as a pivot for the lifted bcs, but something that's scaled better so that it doesn't mess up the strong-connection criterion with GAMG+blocksize
- make sure that vectorial nullspaces are orthonormal (and not just normalized)
- diagnostic solver output: project nullspace out of preconditioned residual in iteration vtus and a new option to dump a vtu with the nullspace vectors

Points for discussion:
- Do we still think velocity reordering is useful? It seems the reordering has only brought (very small) increases in number of iterations. The original strong motivation for it was that with blocksize gamg we would avoid the zero-pivot problem. However in 3D I no longer see why this would be the case. In 2d, did we indeed see zero-pivot problems disappear with this change? In 3D do we still have cases with zero-pivots?
- Does the pivot scaling need further tweaking?
- what to do with commit 479312b?


drhodrid:
Hi Steph,

Thanks for putting this up. In relation to the specific discussion points:

1. From my memory, we did indeed see that the zero pivot situation improved in 2D, with the reordering. I've not recently had a 3D case with a zero pivot issue, but I plan on running a more extensive suite of benchmark cases in the coming weeks and can report back. 
2. Does pivot scaling need further tweaking? I have no idea! Perhaps you can comment (on your own pull request!)? What are your thoughts?
3. Commit 479312b - I would just squash it and tell @drhodrid off for his silly commit message. On a serious note, limiting the coarse equation limit to 800 was recommended by the petsc-devs - do you think that line should be kept or do you not see a motivation for it? I seem to recall it only helped in limited circumstances (although I have no recollection of what they were!).

Perhaps @cianwilson has some more useful comments! I note that the buildbot for this branch is now also green:

http://buildbot-ocean.ese.ic.ac.uk:8080/builders/bc-pivot-scaling

Rhod

cianwilson:
No more useful comments I'm afraid.  Though it is great to see this all in one place as I always struggle to remember what we changed to make gamg behave well.

Perhaps a test would be useful then we can both monitor this going forwards (as gamg/petsc seems to change a lot) and play with turning on and off some of these features for the purposes of deciding what needs to be included.

stephankramer:
Reopening this discussion to propose a merge in the near future.

About the points I raised earlier:
1) I still don't quite understand why this doesn't happen in 3D with a 6-dim near null space, but that's not really relevant for this merge. If anything it should make this potential issue occur less.
2) Looking at it again, I think it's fine and I'll leave it to a capable reviewer to decide.
3) Will fix and tell @drhodrid off for his silliness.

stephankramer:
Pinging @cianwilson for review...

femtools/Solvers.F90
```diff
@@ -1953,6 +1953,8 @@ recursive subroutine setup_pc_from_options(pc, pmat, option_path, &
         ! we think this is a more useful default - the default value of 0.0
         ! causes spurious "unsymmetric" failures as well
         call PCGAMGSetThreshold(pc, 0.01, ierr)
+        ! this was the old default:
+        call PCGAMGSetCoarseEqLim(pc, 800, ierr)
```
cianwilson: Would be good to  have a better explanation for why this is the "right" number.  Is it worth adding it as an option or doing more tests?

cianwilson:
Everything looks good (I realize the line I picked on might not have a good explanation and be just "what worked").

I seem to remember there being an issue with long tests.  Would it be worth it to run through them before a merge of such a large change?

stephankramer:
Thanks @cianwilson 

Yes, it's possible we should have some options indeed to select what the coarse grid solver is (and related to that what the maximum size of this coarse grid is) - as it may be dependent on the problem, e.g. whether it has a (true) nullspace or not, serial vs. parallel, etc. But that's probably for another PR, after some more testing of the various options (redundant svd for example). For the moment I propose to keep that one line in, as it's what's known to work with a lot of cases we care about, and simply reverts a default that was there in petsc <3.6.

@drhodrid is running some long tests for me at the moment. I was mostly worried about dg parallel, as I seem to recall this was an issue previously, and isn't covered in short/medium.

drhodrid:
@stephankramer - Good news - the long tests that we discussed all pass. With that in mind, provided @cianwilson is happy with the code, I'm happy for this to be merge. You're a star.

drhodrid:
OK. I have run the entire long test suite on this branch and master. There are no failures on this branch that do not occur in master, which implies that this branch can now be merged. However, for our records, the following tests are failing consistently on both BC-Pivot-Scaling and Master and, hence, these tests will need looking into post merge (I'm already looking into some, but some have been failing for a very long time it seems):

1. Hokkaido-nansei-oki_tsunami.xml: F ---->>>>> *** FLUIDITY ERROR ***; Source location: (Halos_Communications.F90,  284); Failed assertion size(real_data, 1) >= max_halo_node(halo)

2. medp1dgp2.xml: F ----->>>>> ERROR (Calendar.cpp): The _from_unit specification contains a syntax error.

3. Backward_facing_step_2d_test.xml: PPFFFP ---->>>>> backward_facing_step_2d_longtest: Running reference reattachment length is about 14.0: backward_facing_step_2d_longtest: failure. backward_facing_step_2d_longtest: Running kepsilon reattachment length is approximately equal to the experimental value of 7.0:

4. stagnant_conical_island.xml: F ----->>>> Running no_nodes_doesnt_increase_a_lot: change no_nodes =  2292.0

5. driven_cavity.xml: PPPFPPPP: ----->>>>> Running Quantity4ConvergesToSecondOrder: 1.72761663.

6. gls-StationPapa.xml: PFPPPP ---->>> Final temperature max (not between 6.5 and 7) == 7.62.

7. explicit-hyperc-shear-adapt.xml: F ---->>> Error in unified mesh

**Whilst these tests fail on master only so will be fixed by this merge:**

1. backward_facing_step_3d_longtest: Running Reattachment length does not change by more than 18%: 0.329516894242.

2. flredecomp_long	(partitioning error)

I do not really understand why this merge fixes backward_facing_step_3d_longtest!

stephankramer:
Excellent - pushing it once more on buildbot, so the medium tests are run again with the last few commits in and will then merge.

PR #110, opened 2016-04-21
Allow strong pressure bcs to be used with stokes.

cianwilson:
Following @stephankramer's example, I'm setting up this pull request as a way of discussing this branch, primarily with "team stokes".

This branch aims to allow strong pressure bcs to be used with Stokes.  It proved not to be useful for the problem I was dealing with but it's still useful functionality to have (and surprising that it didn't already work).

What this branch needs is testing though.  Does anyone have a suggestion of a test where strong pressure bcs would be useful?  Perhaps one where we currently use weak bcs (keeping in mind the subtle differences between strong and weak pressure bcs).

### DIFF
--- a/assemble/Full_Projection.F90
+++ b/assemble/Full_Projection.F90
@@ -64,7 +64,7 @@
     
 !--------------------------------------------------------------------------------------------------------------------
     subroutine petsc_solve_full_projection(x,ctp_m,inner_m,ct_m,rhs,pmat,&
-      state, inner_mesh, auxiliary_matrix)
+      state, inner_mesh, auxiliary_matrix, inactive_mask)
 !--------------------------------------------------------------------------------------------------------------------
 
       ! Solve Schur complement problem the nice way (using petsc) !!
@@ -84,6 +84,7 @@
       type(mesh_type), intent(in):: inner_mesh
       ! p1-p1 stabilization matrix or free surface terms:
       type(csr_matrix), optional, intent(in) :: auxiliary_matrix
+      logical, dimension(:), intent(in), optional :: inactive_mask
 
       KSP ksp ! Object type for outer solve (i.e. A * delta_p = rhs)
       Mat A ! PETSc Schur complement matrix (i.e. G^t*m^-1*G) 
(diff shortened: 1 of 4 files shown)


cianwilson:
Buildbot: http://buildbot-ocean.ese.ic.ac.uk:8080/waterfall?show=stokes-strong-pressure-bcs

stephankramer:
Some remarks:
* Would it be possible to move this inactive_mask stuff inside petsc_solve_full_projection() so as not further clobber momentum_equation with it?
* Depends on the answer to the previous question: with petsc_solve_full_projection() I've beaten @drhodrid in submission to comment each subroutine argument - it would be nice if you could follow his shining example :) It's not clear what an `inactive_mask` is here.
* Similarly, when copy and pasting the headers of set_reference_node and apply_dirichlet_conditions please adjust the comments to the specific instance (I know this isn't done in a lot of places)

Did you try this with inhomogeneous pressure bcs as well? I'm not a 100% sure it'll do the right thing by just having the "answer" in the rhs. Maybe we can do a simple Poisseule flow with a nonzero (but constant) pressure bc on the outflow?

cianwilson:
Thanks @stephankramer ...

* so the reason I didn't move inactive_mask into the full projection solver is because you would have to pass in some other stuff, like dt, theta_pg, theta_divergence, the positions to be able to call the boundary conditions routines.  I thought just passing in the inactive_mask was a better option (in fact I took the inactive mask out of full projection for this reason).
* Re: commenting the argument - yeah, sure.  Thanks for keeping me on the straight and narrow.
* And sorry, thought I had been adjusting the header comments, must have missed some.

I did try this with an inhomogeneous pressure, yes, that's why I needed it but perhaps what you're saying is why it didn't work :)  Having said that, I don't think I'm just applying the answer in the rhs.  It is scaled as it would be if we weren't using the full projection solver, i.e. by those dt, theta_pg and theta_divergence arguments I mentioned above.

Your test suggestion sounds good.  Perhaps that will clear up whether things are scaled correctly or not.

drhodrid:
Would it be of use for me to setup and try this test? If so, I may need a quick offline chat with Stephan beforehand... just so that I fully understand the setup he is suggesting!! Very rarely have I been used as a shinning example - but it makes me a happy man ;-)

cianwilson:
I already tried it and it didn't work.  I'll try to dig out why so we can discuss.

Issue #109, opened 2016-04-20
Incorrect kind definition in femtools/Global_Parameters.F90


jrmaddison:
The kinds int_4, int_8, int_16, real_4, real_8, and real_16 are incorrectly defined. For example

``
int_16 = selected_int_kind(16)
``

yields an 8 byte integer

jrper:
I believe selected_int_kind should always return a compiler default integer, in order to be used as a kind parameter. See the definition at e.g.  https://software.intel.com/en-us/node/526741.

jrper:
Although thinking about it, that implies that the actual variables defined using these parameters are likely to be far bigger than they were intended to be.

jrmaddison:
It's a valid kind parameter, but not necessarily the one that is expected (see e.g. the comment around the real_4 definition).

The following should lead to the correct behaviour (up to the platform dependent definition of C types).

```
index 8cf177a..39aa90d 100644
--- a/femtools/Diagnostic_variables.F90
+++ b/femtools/Diagnostic_variables.F90
@@ -31,7 +31,7 @@ module diagnostic_variables
   !!< A module to calculate and output diagnostics. This replaces the .s file.
   use fldebug 
   use global_parameters, only:FIELD_NAME_LEN,OPTION_PATH_LEN, &
-& PYTHON_FUNC_LEN, int_16, integer_size, real_size
+& PYTHON_FUNC_LEN, int_8, integer_size, real_size
   use quadrature
   use futils
   use elements
@@ -174,7 +174,7 @@ module diagnostic_variables
     
     !! Recording wall time since the system start
     integer :: current_count, count_rate, count_max
-    integer(kind = int_16) :: elapsed_count
+    integer(kind = int_8) :: elapsed_count
   end type stat_type
 
   type(stat_type), save, target :: default_stat
diff --git a/femtools/Global_Parameters.F90 b/femtools/Global_Parameters.F90
index f0fc09a..1d31f76 100644
--- a/femtools/Global_Parameters.F90
+++ b/femtools/Global_Parameters.F90
@@ -63,16 +63,13 @@ module global_parameters
   integer, parameter :: real_size=4
 #endif
 
-  integer, parameter :: int_4 = selected_int_kind(4), &
-                      & int_8 = selected_int_kind(8), &
-                     & int_16 = selected_int_kind(16)
+  integer, parameter :: int_8 = c_long
   ! real variable declarations of the form:
   !   real*4 :: real_var
[...]

PR #111, opened 2016-04-25
Use iso_c_binding kind names.

jrper:
This changeset attempts to deal with the issue raised in #109 by explicitly replacing the kind parameters with those from iso_c_binding in all cases.

### DIFF
--- a/femtools/Diagnostic_variables.F90
+++ b/femtools/Diagnostic_variables.F90
@@ -31,7 +31,7 @@ module diagnostic_variables
   !!< A module to calculate and output diagnostics. This replaces the .s file.
   use fldebug 
   use global_parameters, only:FIELD_NAME_LEN,OPTION_PATH_LEN, &
-& PYTHON_FUNC_LEN, int_16, integer_size, real_size
+& PYTHON_FUNC_LEN, integer_size, real_size
   use quadrature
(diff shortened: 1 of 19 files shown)


femtools/Diagnostic_variables.F90
```diff
@@ -31,7 +31,7 @@ module diagnostic_variables
   !!< A module to calculate and output diagnostics. This replaces the .s file.
   use fldebug 
   use global_parameters, only:FIELD_NAME_LEN,OPTION_PATH_LEN, &
-& PYTHON_FUNC_LEN, int_16, integer_size, real_size
+& PYTHON_FUNC_LEN, integer_size, real_size
```
stephankramer: Does this one need a `use iso_c_binding` as well? I'm guessing it currently inherits it from somewhere...

stephankramer:
Great! Thanks a lot, @jrper

jrper:
I've added the extra statement @stephankramer asked for. Could this get a builedbot queue? Not that I'm paranoid or anything.

stephankramer:
Buildbot queue here: http://buildbot.ese.ic.ac.uk:8080/waterfall?show=c_types_fix

After that: ship it!

PR #112, opened 2016-04-27
Enable petsc 3.7 build and cleanup of legacy petsc support

stephankramer:
Changes in petsc 3.7 API:
* every PetscOptionsXXX() call needs an additional PetscOptions context as its
first argument (NULL to refer to the global database). As before
petsc<3.7 is supported via #define macros that drop this first argument.
* the standard monitors as KSPMonitorDefault and KSPMonitorTrueResidual
need a PetscViewerAndFormat context and need a destructor routine
(PetscViewerAndFormatDestroy) when setting them in KSPMonitorSet

Clean up of petsc legacy support:
* Removed all logic associated with
petsc versions before 3.4. This includes removing the hacky
MatCreate[Seq|MPI][B]AIJ wrappers. Now calling MatCreate[B]AIJ directly
and calling MatSetup after. Have not /deliberately/ broken petsc 3.4 or
3.5 support - but only 3.6 and 3.7 is supported (not sure if 3.4 or
3.5 still build).
* removed solvers/diagnostics/monitors/ option 'preconditioned_residual_graph'
which hasn't been supported since petsc 3.4. If someone wants to
reinstate it (it draws a graph of convergence in a seperate X-window),
feel free to do so.


drhodrid:
Hi Steph,

I built this branch today with the petsc 3.7 (tarball) and Zoltan 3.81. The build was fine. However, I had the following (medium) test failures:

stokes_square_convection.xml: F
diamond_validation.xml: F
Stokes_square_convection_1e4_p2p1.xml: F

Both stokes tests are failing the first  pressure solve with an indefinite PC issue. They are currently configured with the outer pressure solve using CG, although we now use FGMRES for this in our standard simulations. 

Do you have any idea why the new petsc would be failing with a CG outer solve? Either way, if we change the outer pressure solve to use FGMRES, all works fine.

Rhod

stephankramer:
Could you try to rerun one of these with CG but with a much tighter tolerance on the inner solve?

drhodrid:
Hi Steph,

Sorry for being slow. I tried the tests with tolerances of 1e-20 on the inner solve. I also tried with a direct inner solve (preonly/lu/mumps). Under both scenarios, the tests fail with an indefinite PC. Thus far, the only thing that seems to get them through is using FGMRES for the outer solve KSP.

Rhod

aclocal.m4
```diff
@@ -505,8 +505,8 @@ implicit none
       one  = 1.0
       neg_one = -1.0
       ione    = 1
-      call PetscOptionsGetInt(PETSC_NULL_CHARACTER,'-m',m,flg,ierr)
-      call PetscOptionsGetInt(PETSC_NULL_CHARACTER,'-n',n,flg,ierr)
+      call PetscOptionsGetInt(PETSC_NULL_OBJECT, PETSC_NULL_CHARACTER,'-m',m,flg,ierr)
+      call PetscOptionsGetInt(PETSC_NULL_OBJECT, PETSC_NULL_CHARACTER,'-n',n,flg,ierr)
```
jrper: Is this going to be safe when running configure for versions of Petsc < 3.7?

jrper:
As far as I can see the code changes all match the PETSc change list, and removing the old legacy support code is sensible. I'm not totally sure that overloading the preprocessor and implicitly making Fortran calls to the PetscOptionsXXX family case sensitive is a brilliant idea (I imagine the error messages people could get might be a little confusing) but I suspect you're the only person likely to touch that region of the code in the near future. I note that this is green on buildbot, so once you're satisfied you understand @drhodrid's issues I think this is good to go.

stephankramer:
Yeah, I agree that it's not a pretty solution. I'm not keen on the alternative of sprinkling #ifdef logic everywhere either. Yet another solution would be to have a wrapper function that is explicitly named differently and has the #ifdef logic inside it - which is more easy to understand - but still more cumbersome to maintain. So for now I'll stick with the current approach.

I'm not sure anyone really understands all of @drhodrid 's issues - this particular one: I have a niggling feeling about it, but will ignore it for now. (I did read some throw away remark on one of the petsc mailing lists about gamg not guaranteeing to be SPD that I should look into) .

For now I'll change these tests to FGMRES and rebase the branch and give it one more run through.

stephankramer:
Oh actually I see none of these actually use GAMG so that's irrelevant. @drhodrid could you confirm that these tests pass on 3.7 for you now? Any idea about that diamond validation failure?

drhodrid:
Hi Steph,

These tests now pass fine using fgmres. I’m not sure I understand the diamond validation error sorry…

Rhod

On 3 Jun 2016, at 01:15, Stephan Kramer <notifications@github.com<mailto:notifications@github.com>> wrote:


Oh actually I see none of these actually use GAMG so that's irrelevant. @drhodrid<https://github.com/drhodrid> could you confirm that these tests pass on 3.7 for you now? Any idea about that diamond validation failure?

tmbgreaves:
Hi Rhod, Steph - any more feedback on this? Can it be merged?

drhodrid:
Aside from the diamond validation failure, I think it's good to go... but I'd be keen to understand that failure first! Any idea what's going on with that Tim?

tmbgreaves:
I'm failing to reproduce the diamond validation error here - do you have any more output which might help debug it, please?

drhodrid:
Ok an update from my end. I tested this branch with petsc 3.7.2 (the latest tarball overnight). All tests passed so I think it's good to merge. However,  the biology_6_component_parallel test is behaving strangely. It passes the test, but the final serial simulation drops out with a seg fault. I tried this on master and I see the same behavior. So somebody needs to look at this test (although this shouldn't hold up this merge). The current test owner is @jhill1. 

Is our test framework designed to pick up segmentation faults?

tmbgreaves:
Thanks Rhod - I don't think we pick up segfaults other than testing the output, good spot.

Issue #117, opened 2016-05-03
Unbuilt executables in the tools directory


jrper:
It's come to my attention that we have broken executables in the tools directory (I rather think I broke them). In particular these are tools not built by the fltools tools target, so they aren't tested or built under buildbot. Is there consensus whether these should be fixed, or removed, and if fixed, should they be built by `make fltools`?

Tools:

- time_average_parallel
- test_laplacian_vector

also there are the Flint_* routines, which don't seem to get compiled into an executable.

stephankramer:
I'd say just get rid off it. Clearly nobody has missed them - and if someone does come along, we can put the effort it then to resuscitate them.

PR #118, opened 2016-05-09
Removing orphaned, untested tools.

jrper:
This changeset applies the work discussed in #117, removing the tests which don't get built with a "make fltools" and which are currently broken. Since this removes file and indirectly affects the build system, it should probably be tested first, so could I have a buildbot queue please?


stephankramer:
Thanks for that, looks good to go.

Buildbot branch here: http://buildbot-ocean.ese.ic.ac.uk:8080/builders/remove_broken_tools

jrper:
Green on buildbot, so merging.

Issue #113, opened 2016-04-27
VTK library name changes


jrper:
As was raised in the AMCG Fluidity developers meeting, the current configure script doesn't support the default namescheme chosen for VTK libraries in (at least) version 6.2 and 6.3, as used in unstable versions of debian. An example of a workaround method is available here: 32eb4909166c6565344fe7d8583927392eb4908c however this is itself incompatible with the centos packages for VTK 6.1. The intention is to extend the existing autoconf scripts to include all reasonable namings.

ggorman:
May well be easier to switch from autoconf to cmake. The learning curve is far easier than autoconf.

stephankramer:
So digging into this some more, this seems to be entirely vtk's fault. vtk indeed now inserts a version number at the end of the lib name /before/ the .so. To suppress this you can add -DVTK_CUSTOM_LIBRARY_SUFFIX='' in vtk's configure, but that of course leaves the package maintainers in the dilemma: if you suppress the ugly non-standard naming scheme, you break third-party software (like us) that expect vtk to be built in the default way. See a.o. this discussion https://bugs.archlinux.org/task/43378. So we should expect different distro's to have different solutions.

As far as I can see there's no way to find out how it's been build, even if we were using cmake. The value of VTK_CUSTOM_LIBRARY_SUFFIX that's been used isn't set in any of the cmake files that come with Debian's build for instance and neither is it in centos' rpm (which does set -DVTK_CUSOM_LIBRARY_SUFFIX=''" in the .spec of the source rpm).

jrper:
That's a nice piece of blackmail then: use cmake as your build system and have code that will work with a script like

    cmake_minimum_required(VERSION 2.6)
    enable_language (Fortran CXX)
    project(bob)
    find_package(VTK REQUIRED)
    include_directories(./include)
    include(${VTK_USE_FILE})
    add_executable(bob bob.cpp bob.F90)
    target_link_libraries(bob ${VTK_LIBRARIES})

Otherwise, write an autoconf script to try every possibility of custom suffices.

ggorman:
And this is why I eventually gave up and started just using cmake for everything.

stephankramer:
[...]

PR #125, opened 2016-07-12
Implementing Stephan's fix for finding VTK.

jrper:
This pull request attempts to implement @stephankramer's cmake based fix for finding where the elves have hidden VTK on an arbitrary system, issue #113. So far it doesn't remove anything, the only time it might change behaviour is when cmake isn't available, but the location of VTK has been implicitly provided through environment flags.

### DIFF
--- a/configure
+++ b/configure
@@ -732,6 +732,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -847,6 +848,7 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
+runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE_TARNAME}'
@@ -1099,6 +1101,15 @@ do
   | -silent | --silent | --silen | --sile | --sil)
     silent=yes ;;
 
+  -runstatedir | --runstatedir | --runstatedi | --runstated \
+  | --runstate | --runstat | --runsta | --runst | --runs \
+  | --run | --ru | --r)
+    ac_prev=runstatedir ;;
+  -runstatedir=* | --runstatedir=* | --runstatedi=* | --runstated=* \
+  | --runstate=* | --runstat=* | --runsta=* | --runst=* | --runs=* \
+  | --run=* | --ru=* | --r=*)
+    runstatedir=$ac_optarg ;;
+
   -sbindir | --sbindir | --sbindi | --sbind | --sbin | --sbi | --sb)
     ac_prev=sbindir ;;
   -sbindir=* | --sbindir=* | --sbindi=* | --sbind=* | --sbin=* \
@@ -1236,7 +1247,7 @@ fi
 for ac_var in	exec_prefix prefix bindir sbindir libexecdir datarootdir \
 		datadir sysconfdir sharedstatedir localstatedir includedir \
 		oldincludedir docdir infodir htmldir dvidir pdfdir psdir \
-		libdir localedir mandir
+		libdir localedir mandir runstatedir
 do
   eval ac_val=\$$ac_var
   # Remove trailing slashes.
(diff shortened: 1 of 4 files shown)


tmbgreaves:
That looks great, thank you. I'll run it through buildbot and attempt to test on various different platforms.

stephankramer:
One query: is `-DCOMPILER_ID=gcc` appropriate when using intel?

jrper:
Experimenting on the college cluster, this appears to give answers of the right form with `-DCOMPILER_ID=gcc` regardless of the compiler which was actually used to compile the vtk library itself. I _think_ that in this case this only controlling the syntax of the output, and -I/vtk/include/stuff and a list of absolute library paths works for both gcc and intel (plus clang). Do we actually claim to support any other compiler?

anguscr:
Dear all,

If I may chip in, last I checked, some work was required to get Fluidity 
to compile with the Intel compilers*.

Regards
Angus

[*] This statement may be completely out of date.


On 13/07/2016 14:51, James Percival wrote:
> Experimenting on the college cluster, this appears to give answers of
> the right form with |-DCOMPILER_ID=gcc| regardless of the compiler which
> was actually used to compile the vtk library itself. I /think/ that in
> this case this only controlling the syntax of the output, and
> -I/vtk/include/stuff and a list of absolute library paths works for both
> gcc and intel (plus clang). Do we actually claim to support any other
> compiler?


-- 
Dr Angus Creech

Research Fellow
Institute of Energy Systems
University of Edinburgh
Scotland

www.research.ed.ac.uk/portal/acreech2

The University of Edinburgh is a charitable body, registered in
Scotland, with registration number SC005336.

tmbgreaves:
@jrper - we currently only claim support for gcc and Intel 2015.

@anguscr - all working nicely for Intel 2015 now :-) 2016 is another story on many counts...

anguscr:
Duly noted, Tim - I do wonder if it's worth the effort to support 2016.

Regards
Angus


On 13/07/2016 15:01, Tim Greaves wrote:
> @jrper <https://github.com/jrper> - we currently only claim support for
> gcc and Intel 2015.
>
> @anguscr <https://github.com/anguscr> - all working nicely for Intel
> 2015 now :-) 2016 is another story on many counts...


-- 
Dr Angus Creech

Research Fellow
Institute of Energy Systems
University of Edinburgh
Scotland

www.research.ed.ac.uk/portal/acreech2

The University of Edinburgh is a charitable body, registered in
Scotland, with registration number SC005336.

stephankramer:
@jrper: Fair enough - I could imagine though that this might depend not only on the specific compiler but also the environment that was used to built VTK, but I'm perfectly happy to go with the assumption this works in general until such time that we do find a platform where asking for the specific compiler seems necessary.

tmbgreaves:
Ubuntu Trusty, CentOS 6, and CentOS 7 tests now running.

tmbgreaves:
The fluidity-dev package in CentOS is missing cmake - I'll push a new version of that and retest.

jrper:
There's an argument that we ought to be testing this in the case that cmake isn't present as well as the case that it is (in fact it doesn't seem to be present in the trusty build on buildbot either, so the new code shouldn't be doing anything apart from adding a few error messages to the logs).

tmbgreaves:
I guess the problem arises that when you have the libraries in non-standard places/names it won't find them unless cmake gives useful information. This feels to me to be a reasonable failure, and expected on CentOS since one of the reasons we started down this route was that CentOS had non-standard VTK library naming -- hence I think it's not unreasonable to require cmake to help out :-)

tmbgreaves:
The RHEL/CentOS builds now have cmake added and are running to test.

jrper:
That's a weird mix of successes and failures we're getting. Are the full config.log files available somewhere?

tmbgreaves:
I'm attempting to replicate them by hand at the moment to get the config out - but getting oddly inconsistent behaviour. I think I've done something interestingly wrong.

tmbgreaves:
There was one minor bug with an additional _ creeping in as CPP_FLAGS (vs CPPFLAGS) which I've fixed and squashed. This builds fine now on CentOS 6, CentOS7 [0], and xenial with vtk5.10, and if one small bug in the libproj package is fixed (softlinking libproj.so) it works with xenial with vtk6.2.

Hence looks good to merge.

[0] The CentOS 7 build is a bit creaky - cmake returns a broken set of include flags for VTK, breaking the automagic (-I//include/vtk instead of -I/usr/include/vtk) but since the subsequent detection looks for the CentOS naming scheme for vtk libraries, the configure and build end up working.

PR #130, opened 2016-09-19
Free surface movement fixes

stephankramer:
This originated as a work branch 'rebecca' for @rcs08. They are all fixes though, that should be merged back in to master:
- "fixing" free surface movement and adaptivity by switching back to the original flat geometry (OriginalCoordinate) just before going into the adapt, and reapplying the free surface movement afterwards
- fix the combination of mesh movement and LinearMomentum
- fix 2+1D adaptivity for the case of having a python expression for maximum edge lengths but not for
  minimum edge lengths

These fixes should probably be tested before merging. Will therefore assign to Branwen when she's been added to the fluidityproject.


stephankramer:
Assigned to BSnelling to add a small test. Maybe you can have a look at one of the standing-wave tests. standing-wave-parallel is already extruded so if you just switch on adaptivity there. Maybe add a volume check (check that the volume is the same throughout the run)....

jrper:
Does this PR need reviewing now, or is it still waiting for tests?

PR #131, opened 2016-10-18
Intersection optimisation

jrper:
This changeset short circuits the vanilla 2d intersection code so that it doesn't pointlessly allocate a vector field when run with "false alarm" elements which touch at a point or edge rather than overlapping. The original version has a significant penalty when running Galerkin projection to interpolate between meshes which are almost identical, and it seems we do something similar by default in 3d anyway.


stephankramer:
Looks good to me.

stephankramer:
Buildbot branch at http://buildbot.ese.ic.ac.uk:8080/waterfall?show=intersection_optimisation

jrper:
Green and merged

PR #134, opened 2016-11-11
Update manual text on interactive shell.

jrper:
The existing fluidity_tools method to get an interactive python shell no longer works with a modern version of IPython (i.e. in the last 4 years or so). Based on an offline discussion with @stephankramer this pulls the failing code out and updates the manual to give a generic version which should work.


stephankramer:
Looks great, thanks! Since this wasn't tested either - I presume we can forgo the usual test cycle?

jrper:
That makes sense to me.

PR #136, opened 2016-11-16
UPSTREAM BUGFIX for libspud

tmbgreaves:
To avoid a destructor being called twice (due to optimisation removing a
variable), deallocated is defined as static in libspud/include/spud and then
defined in libspud/src/spud.cpp.

Fix supplied by James Percival (jrper)

### DIFF
--- a/libspud/include/spud
+++ b/libspud/include/spud
@@ -406,7 +406,7 @@ namespace Spud{
           
       };
       
-      bool deallocated;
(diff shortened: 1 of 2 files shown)


tmbgreaves:
Green on buildbot and fulfilling the requirements for a direct push - a bugfix not touching the Fluidity core core.

PR #138, opened 2016-11-29
Add mesh quality functions from vtk as fluidity diagnostics.

jrper:
This changeset allows the user to specify a mesh quality function from VTK as a fluidity diagnostic field. The intention is that this can be used to control h- and connectivity based mesh adaptivity in the context of moving meshes.

### DIFF
--- a/diagnostics/Makefile.dependencies
+++ b/diagnostics/Makefile.dependencies
@@ -60,7 +60,7 @@ Mesh_Diagnostics.o ../include/mesh_diagnostics.mod: Mesh_Diagnostics.F90 \
    ../include/diagnostic_source_fields.mod ../include/fdebug.h \
    ../include/field_options.mod ../include/fields.mod ../include/fldebug.mod \
    ../include/global_parameters.mod ../include/halos_numbering.mod \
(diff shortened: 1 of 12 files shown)


stephankramer:
This looks great and useful.

Two queries:
* does this work in parallel? I presume it just sets up individual vtk unstructured meshes of the local domains independently, and then it should just work?
* I think this will not work for super-parametric (i.e. CoordinateMesh is not linear). I don't have an objection to that (as my head hurts thinking about which coordinate field it should be using)- but an error message would be nice.

jrper:
1. As you say parallel "just works" for all the functions I've exposed, since they only really require the coordinates of the vertices of the elements that are in the local mesh, and this information is available locally, hence it doesn't even need a halo update. I believe this to be true for all the other VTK mesh quality functions too, but they may not be safe for simplices.

2. That's a good point. Since it's only coded for linear simplices, I'll throw an error if the coordinate is on any other mesh, and add a clearer warning in diamond. 

2.

stephankramer:
http://buildbot.ese.ic.ac.uk:8080/builders/vtk_mesh_quality

PR #142, opened 2016-12-14
Fix to make clean in libspud for CX1

jmn114:
Hi Tim,

James and I spotted a problem when cleaning the libspud library in CX1.

We added the abspath to the SPUDLIB so it returns an absolute name of the library without getting confuse with the CX1 path names.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -372,7 +372,7 @@ clean: clean-light clean-test python_clean clean-debian
 	@cd libvtkfortran; $(MAKE) clean
 	@echo "    CLEAN libjudy"
 	@cd libjudy; $(MAKE) clean >> make.log 2>&1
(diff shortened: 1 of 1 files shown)


tmbgreaves:
Agree with James on the fix needed and happy for this to go in once that's done.

stephankramer:
What did we decide again on how we're going to deal with maintaining libspud/diamond? Move it to git? Was I going to contact @cianwilson about it?

tmbgreaves:
Hi @stephankramer - after some discussion at the dev meeting yesterday could you ask @cianwilson if he's interested in continuing to maintain libspud/diamond on launchpad If not, we were debating between cloning the repository over to Fluidity on github and having a local maintained version, or just going with the version already in the Fluidity tree which is where fixes have been going recently anyway. If you and/or Cian have any input/thoughts there please do post them :-)

PR #143, opened 2017-01-23
Mesh smoothing tm

jrper:
I'm opening this pull request now, so that this doesn't get forgotten about. It will need @tmmcmanus to sign off that he's happy with this as the final version. This is mostly orthogonal to other changes, with the possible exception of the manual changes.


jrper:
TimM has given his blessing, so I'm counting this as open to review. I've asked @yeagerba to look at the mesh movement part, but I suspect @stephankramer may want to comment on the code structure as well.

PR #146, opened 2017-02-12
Hook up the GMRES restart option (and make setting it optional).

jrper:
This changeset means that the GMRES restart option in diamond is used in the code if set. If the option isn't present in the options file then the default value of 30 is used (as always occurs currently).

### DIFF
--- a/femtools/Solvers.F90
+++ b/femtools/Solvers.F90
@@ -1578,7 +1578,7 @@ subroutine SetupKSP(ksp, mat, pmat, solver_option_path, parallel, &
     KSPType ksptype
     PC pc
     PetscReal rtol, atol, dtol
-    PetscInt max_its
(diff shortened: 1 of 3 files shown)


jrper:
In @stephankramer's continued absence, are there any other volunteers to review this?

stephankramer:
Sorry, I missed that! Looks perfect to me. Thanks!

jrper:
Thanks Stephan! @tmbgreaves, can I have a buildbot queue for this? It's theoretically a do-nothing change, but it's best to check.

stephankramer:
There you go:

http://buildbot.ese.ic.ac.uk:8080/waterfall?show=gmres_restart

jrper:
Thanks again. We're green, so I'm merging.

PR #147, opened 2017-02-28
Add name of solve to CPU time and Mflop output, such that different s…

drhodrid:
…olves can be more easily identified in the log file.

### DIFF
--- a/femtools/Solvers.F90
+++ b/femtools/Solvers.F90
@@ -1236,9 +1236,9 @@ logical, optional, intent(in):: nomatrixdump
   if (timing) then
     call cpu_time(time2)
     call PetscGetFlops(flops2, ierr)
-    ewrite(2,*) 'CPU time spent in solver:', time2-time1
-    ewrite(2,*) 'MFlops counted by Petsc:', (flops2-flops1)/1e6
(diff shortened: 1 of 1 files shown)


drhodrid:
Increased verbosity for solver output (CPU Time spent in solves etc...).

jrper:
My only critique here is that the name is that of the associated scalar(vector/tensor) solution field, which might be a sub optimal tag for individual linear solves. I'm not sure I know a better way to organise it though.

drhodrid:
Hi James,

Thanks for looking at it. I maintained the naming convention that was already used throughout that routine (n/o iterations/ksp converged reason etc…) for consistency. It’s clear enough for what I need.

Best wishes,

Rhod

On 1 Mar 2017, at 01:22, James Percival <notifications@github.com<mailto:notifications@github.com>> wrote:


My only critique here is that the name is that of the associated scalar(vector/tensor) solution field, which might be a sub optimal tag for individual linear solves. I'm not sure I know a better way to organise it though.

cianwilson:
Looks good to me.

drhodrid:
Thanks both. All merged.

PR #148, opened 2017-03-05
Gnomonic

stephankramer:
This PR proposes to change the vertical extrapolation code, which extrapolates solutions on a surface mesh vertically to the interior of the entire mesh, for the case of spherical geometries. 

Currently this uses a stereographic projection of the two hemispheres. The downside of this approach is that the stereographic projection does not map straight lines to straight lines. In fact, neither the straight edge between two vertices on a sphere, nor the great circle (segment) over the sphere map to straight lines in the projection.

In a gnomonic projection however greater circles are, by definition mapped to straight lines. This can be achieved by mapping all points to planes at unit distance of the origin. A segment of a greater circle between two surface nodes, and the corresponding straight edge are then mapped to the same line segment on the plane.

Having the edges of the surface mesh map to straight edges in the projected space, improves the accuracy of the vertical extrapolation. Previously a node in the interior was mapped to the same projected space, and found in the projected surface mesh using pickers. Because the projected, now curved edges of the surface mesh are necessarily approximated by straight lines however, it was possible for points that were inside the original, curved projected triangle, to end up on the outside and therefore be interpolated from a neighbouring triangle. As an example, an edge of the surface mesh on the equator is mapped through the stereographic projection to the unit circle; in the projected surface mesh however, these curved circle segments are again approximated by line segments. When searching nodes below such an edge, it is however still mapped onto the unit circle and therefore ends up on one side of the edge in the projected surface mesh.

So far the accuracy of this vertical extrapolation has not been very important, as it was only used to vertically extrapolate the diagnostic free surface downwards - and for mesh movement with a free surface, but for simulations on the sphere this non-linearity has not been realistically resovled. I'm planning on using this improved mapping in spherical adaptivity however.

### DIFF
--- a/femtools/Vertical_Extrapolation.F90
+++ b/femtools/Vertical_Extrapolation.F90
@@ -38,7 +38,7 @@ use global_parameters, only: FIELD_NAME_LEN
 use elements
 use parallel_tools
 use spud
-use integer_set_module
+use data_structures
 use sparse_tools
 use transform_elements
 use fetools
@@ -80,6 +80,8 @@ public vertical_element_ordering
 public VerticalProlongationOperator
 public vertical_extrapolation_module_check_options
 public compute_face_normal_gravity
+
+  real, parameter :: GNOMONIC_BOX_WIDTH = 10.0
 
 contains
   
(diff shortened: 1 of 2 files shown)


cianwilson:
Following an offline discussion this looks good to me!

stephankramer:
Merged

PR #153, opened 2017-03-31
This branch adds a halo update to the inverse lumped mass matrix used in the bound (diffuse) operations. This seems to be needed on the Raijin HPC here in Canberra.

drhodrid:


cianwilson:
Just to clarify, the lumped mass was only being assembled in elements where at least one node was owned by the process you were on.  This lead to 0s in the lumped mass, which was then inverted and used in the bounded diffusion algorithm, leading to NaNs in the halos of the deviation field.  This fixes that.

cianwilson:
Rhodri:  have tests been run on this?

drhodrid:
Hi Cian,

Yes - both the short and medium test suites pass.

Rhod

cianwilson:
Looks good then.

PR #155, opened 2017-04-15
Remove spurious sends

stephankramer:
This aptly named branch removes spurious sends and recvs from the subroutine remove_spurious_sends() which is called from make_global_numbering whenever a mesh is derived with different polynomial degree or continuity. The current implementation calls MPI_ISend and MPI_RECV for every other process which causes issues on jobs with a large number of processes. This was seen on [Raijin](http://nci.org.au/systems-services/peak-system/raijin/) for nprocs>512. 

Regardless of whether a HPC platform should be able to handle this, this practice is not efficient as the very many zero-sized messages do not come for free. Therefore here we change this to only communicate with neighbouring processes in the halo.

A similar pattern was found in expand_halo (called to restore the L2 halo after an adapt). Here avoiding unnecessary communication proved a little more tricky as the halo is at that point still being constructed so neighbourship relations based on the halo still need to be estabilisged.

Finally this pattern has been fixed in the detectors code where detectors move into the halo and are therefore sent to their owners.

I believe this fixes all the places where MPI_(I)Send and MPI_(I)Recv is called from/to every process. There are two remaining cases in the code that look problematic (just mentioning here for documentation sake - not going to be fixed in this PR)
- in the detectors code if on any process a detector leaps outside the halo (and therefore the local processor no longer know where it should be sent to) a special communication is performed that involves N * NPROCS MPI_BCAST calls. Here N is the maximum number of "lost" detectors on any process. I'm not sure how often this actually happens or whether we can avoid it: if a detector moves out of the halo within a timestep, surely we should be able to break off its trajectory and pass it over to a new owner to continue the movement within that timesteps (who can then send it onwards to yet another process)?
- a similar pattern with NPROCS MPI_BCAST calls happens in the zoltan redestribution of detectors. The current strategy seems to involve not bothering zoltan at all with the detectors and then when all is done hoping that the detectors that we used to own are still within our domain - which is absurd - if it no longer is, we invoke the expensive NPROCS MPI_BCAST pattern. This presumably could be fixed by packing the detector information to let zoltan move the detector from the current element owner to the new one.

### DIFF
--- a/femtools/Detector_Parallel.F90
+++ b/femtools/Detector_Parallel.F90
@@ -36,9 +36,9 @@ module detector_parallel
   use elements
   use parallel_tools
   use parallel_fields
-  use halos
   use fields
   use state_module
+  use halos
   use detector_data_types
   use detector_tools
   use pickers
--- a/femtools/Detector_Parallel.F90
+++ b/femtools/Detector_Parallel.F90
@@ -359,7 +359,7 @@ contains
        ndet_to_send=send_list_array(target_proc)%length
        ! if we don't know any elements owned by target_proc - we shouldn't have anything to send
        if (halo_receive_count(ele_halo, target_proc)==0) then
-         ! check that this is te case
+         ! check that this is the case
          if (ndet_to_send>0) then
            FLAbort('send_list_array should only send detectors to known elements.')
          end if
--- a/femtools/Detector_Parallel.F90
+++ b/femtools/Detector_Parallel.F90
@@ -34,10 +34,9 @@ module detector_parallel
   use integer_hash_table_module
   use mpi_interfaces
   use elements
-  use halo_data_types
   use parallel_tools
-  use halos_numbering
   use parallel_fields
+  use halos
   use fields
   use state_module
   use detector_data_types
@@ -309,6 +308,8 @@ contains
     ! receives serialised detectors from all procs and unpacks them.
     type(state_type), intent(in) :: state
     type(detector_linked_list), intent(inout) :: detector_list
+    ! the assumption is here that we only send detectors located in element that we know about
+    ! in the largest element halo
     type(detector_linked_list), dimension(:), intent(inout) :: send_list_array
 
     type(detector_buffer), dimension(:), allocatable :: send_buffer, recv_buffer
@@ -332,6 +333,7 @@ contains
     xfield => extract_vector_field(state,"Coordinate")
     dim=xfield%dim
     allocate( sendRequest(nprocs) )
+    sendRequest = MPI_REQUEST_NULL
 
     ! Get the element halo 
     halo_level = element_halo_count(xfield%mesh)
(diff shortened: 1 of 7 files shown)


stephankramer:
@drhodrid to run the following long tests before merging:

backward_facing_step_3d_longtest
coarse-corner
flredecomp_long
lituya_bay_parallel_debug
lock_exchange_3d_parallel
mphase_tephra_settling_2d_adaptive
mphase_tephra_settling_3d
mphase_tephra_settling_salinity_adaptive
open_ocean_deep_convection-parallel
para_eddy
pseudo2dadapt_parallel_galerkinproj
saltfinger2d_adaptive
sphere-3D-drag-Re100
sphere-3D-drag-Re1000
stagnant_conical_island
viscosity_2d_p0_adaptive_parallel
wetting_and_drying_balzano1_dg_parallel
wetting_and_drying_thacker_dg_parallel

drhodrid:
All test running. I'll update you in a few days.

drhodrid:
All tests pass. However, there seems to be an issue with coarse-corner (unrelated to this merge), where it is failing in the following line of the .xml:

cp -r ../$(MODEL)_*6_checkpoint* .

I cannot see where $(MODEL) is being defined. Accordingly, I had to make this copy manually to verify that the results were still fine.

drhodrid:
I just committed a fix for the issue with the coarse corner test. With that in mind, this branch is good to go, pending a review! @jrper - fancy taking a look? Oh go on!!! ;-)

femtools/Detector_Parallel.F90
```diff
@@ -355,6 +357,14 @@ subroutine exchange_detectors(state, detector_list, send_list_array)
     allocate(send_buffer(nprocs))
     do target_proc=1, nprocs
        ndet_to_send=send_list_array(target_proc)%length
+       ! if we don't know any elements owned by target_proc - we shouldn't have anything to send
+       if (halo_receive_count(ele_halo, target_proc)==0) then
+         ! check that this is te case
```
jrper: Typo in comment.

stephankramer:
For future reference: yes, we discussed this previously off line; The idea is that you could basically renumber the halos such that you'd only loop over the non-empty halos rather than all processes in the communicator. So you'd have something like:

    do i=1, halo_neighbour_count(halo)
      ! do stuff based on halo_sends(halo, i) or halo_receives(halo,i)
      rank = halo_neighbour_rank(halo, i)
      call MPI_IRECV/SEND(...., rank, MPI_COMM_FEMTOOLS)
    end do

You would probably also want to change `node_owner` and `element_owner` to return the local neighbour number of the owner rather than its rank (and return 0 for "owned" nodes/elements).

The rationale for this change would be that with highly parallel runs (1000s of cores) the number of cores becomes of the same order as the number of DOFs per core. So having all these loops over all processes, as we currently have, will at some point limit scalability.

Of course this would be a very intrusive change, having to change any code that does anything with halos (other than simply calling halo_update), with the potential for many bugs. So I'd say this effort is only warranted if we can actually demonstrate this to be a problem (premature optimisation...)

stephankramer:
Merged

PR #156, opened 2017-05-04
Bc smoothing

stephankramer:
Adds an option to smooth the values applied as a boundary condition. The smoothing is done via a simple masslumped projection - higher order functions are evaluated on a P1 mesh, then smoothed, then mapped to the bc value field. Currently only implemented for the components of vector bcs, but could easily be extended to scalar bcs.

Includes test for new functionality (testing monotonicity of a step valued bc)

Includes fix for transform_to_physical on embedde mesh meshes (mesh_dim<positions%dim)

### DIFF
--- a/assemble/Free_Surface.F90
+++ b/assemble/Free_Surface.F90
@@ -1067,7 +1067,7 @@ contains
    call get_boundary_condition(fs, "_implicit_free_surface", surface_mesh=surface_mesh, &
      surface_node_list=surface_node_list)
    if (IsParallel()) then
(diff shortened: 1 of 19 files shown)


stephankramer:
Also moves and renames generate_free_surface_mesh_halos() to generate_surface_mesh_halos() so it can be used for other purposes.

Also, adds a optional argument to tell halo_update to be quiet - which is useful when a series of halo_updates happen inside a loop.

cianwilson:
Looks great!

Nice recent (major) bug fix!

PR #157, opened 2017-05-05
Make schema spud base language compliant for flredecomp zoltan option

drhodrid:
The schema currently isn't spud base language compliant when setting the partitioner option for zoltan (graph|hypergraph) under flredecomp. 

This commit fixes that and updates the only test that currently uses this option. Thanks to Cian for highlighting the issue.

Buildbot branch here:
http://buildbot-ocean.ese.ic.ac.uk:8080/builders/zoltan-option-schema-fix


drhodrid:
Thanks Steph. All merged.

PR #159, opened 2017-06-16
Fluidity/fix/med tides

tmbgreaves:
Fixes to the Mediterranean Tides example, plus removal of Fortran interface that was causing data corruption.


tmbgreaves:
Buildbot running at http://buildbot.ese.ic.ac.uk:8080/builders/Fluidity%2Ffix%2FmedTides

PR #163, opened 2017-08-07
Ksp caching

stephankramer:
This PR does two things:

* it fixes the cache_solver_context option under the solver/ options of  scalar prognostic fields. It was creating a new ksp regardless of the option, so the only thing it did was create a memory leak by not destroying the previous one. This is now fixed. It's still a bit of a feeble option as it relies on users to know whether a matrix changes between subsequent solves.

* it introduces reusing of the ksp for momentum solves. This only applies for full_momentum or mass_matrix solves where there are in fact 3 different places that require setting up of a ksp for a momentum solve:

1) the initial velocity solve
2) the "inner" velocity solve of the Schur complement solve in the pressure projection
3) the velocity correction

Solver options for 1) are directly under Velocity/prognostic/solver whereas solver options for both 2) and 3) are in Pressure/prognostic/scheme/use_projection_method/full_schur_complement/inner_matrix/solver. For full_momentum 1), 2) and 3) in fact solve the same system - for the (consistent) mass_matrix approach 1) is the full momentum matrix and 2) and 3) are a mass matrix solve. What this PR changes is, that it reuses the ksp where possible. For full_momentum this means the ksp setup time (which can be very large when solving a Stokes system with AMG) is divided by three. This reuse is always done automatically without any option. It does not reuse the ksp between, nonlinear iterations or timesteps. 

To accommodate people wanting to use different solver settings in 1) from those used in 2) and 3), the setting of solver options of 2) and 3) under inner_matrix is now optional. If these are not specified, the same solver options are used for both (and for full_momentum the same ksp is used) - if they are specified, we still reuse the ksp (if possible) but set its options again and PETSc should be clever enough to figure out whether it can reuse the preconditioner in that case (e.g. only setting a tighter ksp tolerance does not require the PC to be setup again, whereas if we choose a different pctype it does).

### DIFF
--- a/assemble/Full_Projection.F90
+++ b/assemble/Full_Projection.F90
@@ -207,7 +207,7 @@
       character(len=OPTION_PATH_LEN) :: inner_option_path, inner_solver_option_path
       integer, dimension(:,:), pointer :: save_gnn2unn
       type(integer_set), dimension(velocity%dim):: boundary_row_set      
-      integer reference_node, stat, i, rotation_stat
+      integer reference_node, i, rotation_stat
       logical parallel, have_auxiliary_matrix, have_preconditioner_matrix
 
       logical :: apply_reference_node, apply_reference_node_from_coordinates, reference_node_owned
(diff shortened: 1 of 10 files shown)


cianwilson:
Looks good.

Buildbot?

tmbgreaves:
Cian Wilson wrote:
> Buildbot?

minor glitch with buildbot at the moment (I've
misconfigured ssh and am off-site..) but will at
the latest sort it out in a couple of weeks when
I'm back on site again.

Tim

drhodrid:
@tmbgreaves - what is the issue with Buildbot? This branch is currently green (ran only a few days ago): http://buildbot.ese.ic.ac.uk:8080/builders/ksp-caching - can we assume that this green is a "true green"?

tmbgreaves:
Sounds good - we just had a glitch stopping making
new builders but if you have one, all is good.

Tim

Rhodri Davies wrote:

PR #164, opened 2017-08-07
Variable external density

stephankramer:
This allows for the external density under the viscous free surface to be space-varying. The option is changed from a simple real to a choice between constant, python or from_field.

### DIFF
--- a/main/Fluids.F90
+++ b/main/Fluids.F90
@@ -731,11 +731,6 @@ contains
           
        ! Call the modern and significantly less satanic version of study
        call write_diagnostics(state, current_time, dt, timestep)
-       ! Work out the domain volume by integrating the water depth function over the surface if using wetting and drying
-       if (have_option("/mesh_adaptivity/mesh_movement/free_surface/wetting_and_drying")) then
-          ewrite(1, *) "Domain volume (\int_{fs} (\eta.-b)n.n_z)): ", calculate_volume_by_surface_integral(state(1))
-       end if 
-
 
        if(have_option("/timestepping/adaptive_timestep")) call calc_cflnumber_field_based_dt(state, dt)
 
--- a/main/Fluids.F90
+++ b/main/Fluids.F90
@@ -731,6 +731,11 @@ contains
           
        ! Call the modern and significantly less satanic version of study
        call write_diagnostics(state, current_time, dt, timestep)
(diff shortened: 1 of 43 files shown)


main/Fluids.F90
```diff
@@ -865,11 +865,6 @@ SUBROUTINE FLUIDS()
           
        ! Call the modern and significantly less satanic version of study
        call write_diagnostics(state, current_time, dt, timestep)
-       ! Work out the domain volume by integrating the water depth function over the surface if using wetting and drying
```
cianwilson: Don't object to this being removed as it's a very random thing to be here but it seems orthogonal to the rest of the pull request.  Worth at least mentioning in the summary?

preprocessor/Boundary_Conditions_From_Options.F90
```diff
@@ -1187,6 +1193,26 @@ subroutine set_vector_boundary_conditions_values(state, field, bc_path, position
...
+           if (have_option(trim(bc_path_i)//"/external_density")) then
+              call get_boundary_condition(field, i+1, surface_mesh=surface_mesh, &
+                 surface_element_list=surface_element_list)
+              scalar_surface_field => extract_scalar_surface_field(field, bc_name, name="ExternalDensity")
+              if (have_option(trim(bc_path_i)//"/external_density/from_field")) then
+                ! from_field: The parent field contains the boundary values that you want to apply to surface_field.
+                call get_option(trim(bc_path_i)//"/external_density/from_field/parent_field_name", parent_field_name)
+                scalar_parent_field => extract_scalar_field(state, parent_field_name, stat)
+                if(stat /= 0) then
+                  FLExit("Could not extract parent field. Check options file?")
```
cianwilson: Perhaps worth adding to the error here "parent *scalar* field" as you've only checked for that (understandably) but the schema documentation (because its been reused from other purposes) is unclear about this.

tmbgreaves:
Cian Wilson wrote:
> Buildbot?

minor glitch with buildbot at the moment (I've
misconfigured ssh and am off-site..) but will at
the latest sort it out in a couple of weeks when
I'm back on site again.

Tim

drhodrid:
Green buildbot here: http://buildbot.ese.ic.ac.uk:8080/builders/variable-external-density

stephankramer:
Rebased on master, implemented changes requested by @cianwilson. Have given buildbot one more spin with this

PR #166, opened 2017-09-20
Incorporate spud as subtree cloned from upstream spud

stephankramer:
Now spud has moved over from launchpad with bzr to github with git under the FluidityProject, we can make the syncing of upstream changes in spud into fluidity a lot easier using git subtree (https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree)

The big advantage of this over other methods, is that it requires no change in workflow to people that just checkout fluidity, or only commit changes outside of the libspud/ directory. If changes to spud are required we should make them upstream first (*), and then use the following command:

    git subtree pull -P libspud/ git@github.com:FluidityProject/spud.git master --squash

Note that the `--squash` is required (I used `--squash` in the original `subtree add` and it seems you can't mix-and-match squash and non-squash pulls.) Instead of `git@github.com:FluidityProject/spud.git ` you can add it as a remote in your fluidity checkout:

  git remote add -f spud git@github.com:FluidityProject/spud.git 

and use just use `spud` instead.

(*) Unfortunately it seems `git subtree push` is not working as intended, so we cannot make a local commit to spud in the fluidity checkout and then `git subtree push` it to push it upstream. The problem is that it will go through fluidity's entire history and push all commits that touch the libspud/ directory, adding a lot of noise in spud's history. So I suggest we just refrain from doing that, and only commit changes upstream *first*, and then pull them into fluidity.


jrper:
This looks good to me and is green on buildbot: http://buildbot.ese.ic.ac.uk:8080/builders/subtree-spud

stephankramer:
Redone on top of master.

libspud/src/tests/test_main.cpp
```diff
@@ -38,12 +38,12 @@ extern "C"{
 
 int main(int argc, char **argv) {
 #ifdef HAVE_MPI
-  MPI_Init(&argc, &argv);
+  MPI::Init(argc, argv);
```
jrper: This reverts the switch away from the C++ mph bindings. Which is fine, we just need to remember to back port the change upstream.

libspud/src/tests/test_main.cpp
```diff
@@ -38,12 +38,12 @@ extern "C"{
...
   chdir(getenv("PWD"));
 #endif
   TESTNAME();
 #ifdef HAVE_MPI
-  MPI_Finalize();
+  MPI::Finalize();
```
jrper: see above.

PR #169, opened 2017-12-06
Poiseuille test case

Patol75:
This branch adds a new test case to Fluidity, which is the classical two-dimensional Poiseuille flow in a rectangular cartesian domain. Fluidity solves for the Pressure and the Velocity. The values of the Velocity are then compared to the well-known analytical solution, which depends on the Pressure Gradient that Fluidity works out as well.

Everything seems to work well, however two remarks :
- the pressure boundary conditions are weakly applied, which requires to turn ON "scalar_field (Pressure)/prognostic/spatial_discretisation/continuous_galerkin/integrate_continuity_by_parts". If it is OFF, Fluidity does not raise any Error or Warning and the calculated Pressure field is zero.
- calculating the Pressure Gradient in Fluidity requires to add a FiniteElementGradient field, however the output field name does not specify which field's gradient has been calculated, which can be annoying when multiple FiniteElementGradient fields are defined.

### DIFF
--- a/tests/Stokes_Poiseuille/Makefile
+++ b/tests/Stokes_Poiseuille/Makefile
@@ -1,0 +1,5 @@
+input: clean
+	gmsh -2 -bin -o mesh/Rectangular_Pipe.msh mesh/Rectangular_Pipe.geo
+
+clean:
+	rm -rf *Poiseuille_* fluidity.* P4* *.pyc *.stat mesh/*.msh
--- a/tests/Stokes_Poiseuille/Setup.py
+++ b/tests/Stokes_Poiseuille/Setup.py
@@ -18,4 +18,4 @@ h = 1.
 # Driving pressure, the pressure applied at the inflow boundary
 drvP = 1e6
 # Pressure at the outflow boundary
-outP = 0.
+outP = 5e5
--- a/tests/Stokes_Poiseuille/Setup.py
+++ b/tests/Stokes_Poiseuille/Setup.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+
+# Setup values for the Poiseuille flow simulation
+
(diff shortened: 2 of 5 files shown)


cianwilson:
Hi Thomas,

Thanks for the new test.

Only minor comments... Is it necessary to have it run on 4 processes?  Generally tests should be as small as possible.  Perhaps for such a simple case it would be possible to reduce the domain size/resolution or just run in serial?  In a similar vein, your .geo file seems quite complicated for what ends up being, as far as I can tell, a rectangular mesh with roughly uniform unstructured resolution.  Maybe this could be simplified too?

As to your other comments... yes, both are (very) annoying "features".  Perhaps raise them as issues?  The pressure gradient problem is because you're using "magic named" fields instead of diagnostic algorithms.  It doesn't look like the finite element gradient has been implemented as the latter yet, though there are some equivalents (grad, finite_element_divergence_transpose).  But it's certainly fixable relatively simply.  The first problem with weak bcs and integrating by parts is a little more involved in various bits of assembly code but also fixable.

Cheers,
Cian

Patol75:
Hi Cian,

Thank you for this reply.

I made the test case run in serial instead of parallel, as suggested, and no loss of speed nor accuracy was noticed. Regarding the .geo file, I am a newcomer to Gmsh and the only knowledge I have comes from the documentation, therefore I am not really able to say if a script looks quite complicated or not. Do you believe I should modify the domain definition or the mesh definition (or both) ?

And indeed thank you for the comments and suggestions in the second paragraph.

Thomas

cianwilson:
Hi Thomas,

Sorry for slow response.  Looks good to me.  Happy to merge it in yourself or would you like me to?

Cheers,
Cian

Patol75:
Hi Cian,

I think it is all done now, thank you for the support on this one.

Thomas

PR #170, opened 2017-12-06
Update vtkfortran for VTK MPI communication.

jrper:
It appears that since about VTK 6.3, their code now uses parallel communication to identify which pieces of a parallel file have been written. See line 213 and onward in their github repository [here](https://github.com/Kitware/VTK/blob/3276be1e1f3d44eb403ebda9ad589869c5b7d4bf/IO/ParallelXML/vtkXMLPDataObjectWriter.cxx). 

The net result on an unlucky setup is that our VTK writers produce a summary .pvtu file referencing only one of the .vtus, even though they have all been written. This changeset attempts to deal with this by doing the necessary work to allow the MPI communication to succeed, but has only been tested locally on one system. Possibly @jmn114 might be able to assist with testing.

### DIFF
--- a/configure
+++ b/configure
@@ -15869,6 +15869,11 @@ cat > include/vtk.h <<EOF
 #include <${vtk_header_relative_path}vtkXMLUnstructuredGridWriter.h>
 #include <${vtk_header_relative_path}vtkZLibDataCompressor.h>
 
+#if VTK_MAJOR_VERSION>6 || (VTK_MAJOR_VERSION ==6 && VTK_MINOR_VERSION >2)
+#include <${vtk_header_relative_path}vtkMPIController.h>
+#include <${vtk_header_relative_path}vtkMPICommunicator.h>
+#endif
+
 #ifndef vtkFloatingPointType
 #define vtkFloatingPointType vtkFloatingPointType
 typedef float vtkFloatingPointType;
--- a/configure.in
+++ b/configure.in
@@ -1693,6 +1693,11 @@ cat > include/vtk.h <<EOF
 #include <${vtk_header_relative_path}vtkXMLUnstructuredGridWriter.h>
 #include <${vtk_header_relative_path}vtkZLibDataCompressor.h>
 
+#if VTK_MAJOR_VERSION>6 || (VTK_MAJOR_VERSION ==6 && VTK_MINOR_VERSION >2)
+#include <${vtk_header_relative_path}vtkMPIController.h>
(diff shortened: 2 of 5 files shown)


stephankramer:
I think you forgot to commit libvtkfortran/configure

jrper:
Yes, fixed. Actually, more embarrassingly, I forgot to run autoconf to create it, having patched libvtkfortran/include/vtk.h by hand.

stephankramer:
OK, did you actually test this then? :-) Cause it looks we're not actually using libvtkfortran/configure.in (we should probably delete it then) and include/vtk.h actually gets written and copied to vtkfortran by fluidity's own configure.in

jrper:
I tested that the c++ changes fixed the problem with VTK7 on my system, and that the code still configured and ran with the default AMCG packages. I may have neglected to test whether it configured correctly with VTK7 ...

stephankramer:
Good to go after green buildbot:

http://buildbot.ese.ic.ac.uk:8080/waterfall?show=VTK7-parallel

jrper:
This is green and approved, so pushing. @tmbgreaves If I succeed in getting a dockerfile and image which builds VTK 8.1, is there any chance of adding it to buildbot?

jrper:
That was the best soothsaying I could come up with, based on commit dates in the history. If you think it's only higher versions, then we may as well just say VTK 7 and higher. Do you want to make the change and commit it as a bug fix? I suppose it's even possible this might be configure flag dependent...

drhodrid:
Hi James,

Sure thing - I’ll do this tomorrow.

Best wishes,

Rhod

---------------------------------------------------------------------------------------------------
ARC Future Fellow
Research School of Earth Sciences,
The Australian National University,
Mills Road, Acton,
Canberra, ACT, 2601, Australia

T: +61 (0)2 6125 3643
E: rhodri.davies@anu.edu.au<mailto:rhodri.davies@anu.edu.au>
W: http://rses.anu.edu.au/people/dr-rhodri-davies
CRICOS Provider No. 00120C
---------------------------------------------------------------------------------------------------


On 1 May 2018, at 6:08 pm, James Percival <notifications@github.com<mailto:notifications@github.com>> wrote:


That was the best soothsaying I could come up with, based on commit dates in the history. If you think it's only higher versions, then we may as well just say VTK 7 and higher. Do you want to make the change and commit it as a bug fix? I suppose it's even possible this might be configure flag dependent...

PR #171, opened 2017-12-07
Remove Magically named DiagnosticCoordinate field and add a replaceme…

drhodrid:
…nt diagnostic algorithm, which allows for various remapping functionality. Modified and tested relevant test cases accordingly.

### DIFF
--- a/assemble/Diagnostic_fields_wrapper.F90
+++ b/assemble/Diagnostic_fields_wrapper.F90
@@ -295,13 +295,6 @@ contains
          end if
        end if
 
-       v_field => extract_vector_field(state(i), "DiagnosticCoordinate", stat)
-       if(stat == 0) then
-         if(recalculate(trim(v_field%option_path))) then
-           call calculate_diagnostic_variable(state(i), "DiagnosticCoordinate", v_field)
-         end if
-       end if
-
        v_field => extract_vector_field(state(i), "BedShearStress", stat)  
        if(stat == 0) then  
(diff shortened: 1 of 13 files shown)


stephankramer:
Moving from "magic name" DiagnosticCoordinate to "new style" diagnostic algorithm is great, thanks for doing that. I do have a query about the fact that you've reimplemented the algorithm rather then using the existing code, which simply calls `get_nodal_coordinate()`. As I understand, this is because the latter doesn't work for the case that:
* the mesh of the diagnostic field doesn't have a Coordinate field of its own (`"SomeMeshCoordinate"`)
* *and* the CoordinateMesh is higher order than that of the diagnostic field

If that's the case, I'd rather we'd fix `get_nodal_coordinate()`, rather than creating a new version which does something slightly different. In particular your new algorithm ignores `"SomeMeshCoordinate"` altogether - which in some cases doesn't do the right thing: for the horizontal mesh with mesh extrusion, and any mesh that's excluded from mesh adaptivity, the coordinate field can't be reconstructed by simply remapping `"Coordinate"`. Periodic (unperiodising and reperiodising, etc.) _might_ do the right thing but not sure - and in any case it would be better to be consistent with what happens in the rest of the code.

Things would probably be simplified a lot if we had an associated `"SomeMeshCoordinate"` field for every `"SomeMesh"` mesh. But that would require some work in populate state: we'd have to make sure that higher order derived meshes (which currently don't have their own coordinate field), are hit with the right superparametric mapping in the right place.

PR #176, opened 2018-02-21
Enhancement of RK detector advection and bug fixing

Cmath2:
A summary of the changes that have been made:

Fixing the RK advection scheme:
-The RK advection scheme was updated to include timestep-nodes in its calculations, this involved several steps
-a timestep_nodes parameter was added in Detector_Data_Types.F90, this was done to enable allocation of these values
-Detector_Move_Lagrangian.F90 had several changes made, this included the allocation of a timestep_nodes array, and the setting of timestep_nodes values
-Detector_Move_Lagrangian.F90 was also edited to include a vfield_old parameter. This was done to enable accurate calculation of RK particle advection, through the use of velocity values at both the current and previous timestep.
-To enable this, Populate_State.F90 was edited to allocate an "OldVelocity" field, based off the current velocity field before it is updated.

Addition of test cases:
-Two time-varying velocity test cases were added (lagrangian_detectors_increasing_rotation and lagrangian_detectors_oscillating_rotation). This was done to test the updated runge-kutta advection scheme used for the advection of detectors.
-get_RK_traj.py was also updated to include timestep node values (as the timestep weights were all previously calculated independent of a time value)
-4th and 2nd order convergence tests were added for the new RK particle advection scheme

Cleanup and Bug fixing:
-adaptivity was removed from time dependent detector test cases
-Populate_State.F90 edited to always have a nonlinear velocity field if vfield is prescribed, this means no specific logic is required inside relax_to_nonlinear for prescribed fields
-the default output type for detectors was changed to binary, with an option to make the output type ascii included in the flml file.
-Edited Diagnostic_variables.F90. Fixed bug for check-pointing single detectors in binary
-Edited Timeloop_utilities.F90. Added relaxation term under temporal discretization for prescribed velocity fields. Set default value of itheta for prescribed fields to be 1.0
-Edited Populate_state.F90. Defined the field type of "OldVelocity" to be the same as the current field.

A buildbot of this branch can also be viewed here 
http://buildbot.ese.ic.ac.uk:8080/builders/chris_detectors_final/builds/0


tools/update_options.py
```diff
@@ -13,7 +13,7 @@
 def Help():
   debug.dprint("Usage: update_options [OPTIONS] ... [FILES]\n" + \
                "\n" + \
-               "Updates flml, bml, swml and adml files. If FILES is not specified, all .flml, .bml, .swml and .adml files in\n" + \
+               "Updates flml, and adml files. If FILES is not specified, all .flml, and .adml files in\n" + \
```
cianwilson: Changes to this file seem somewhat tangential to this pull request.

Why were these changes made?

tests/lagrangian_detectors_increasing_rotation/plotscript.py
```diff
@@ -0,0 +1,21 @@
+from scipy import *
```
cianwilson: Add a comment to say what this script does.  Doesn't appear to actually plot anything, despite name. Necessary?  Seems to overlap with analytical_traj.py.

tests/lagrangian_detectors_increasing_rotation/analytical_traj.py
```diff
@@ -0,0 +1,24 @@
+from pylab import *
```
cianwilson: Add a comment saying what this script does.  Appears to both generate Xvals and Yvals and open a plot.

tests/lagrangian_detectors_oscillating_rotation/plotscript.py
```diff
@@ -0,0 +1,21 @@
+from scipy import *
```
cianwilson: Add a comment to say what this script does. Doesn't appear to actually plot anything, despite name.

tests/lagrangian_detectors_oscillating_rotation/plotscript.py
```diff
@@ -0,0 +1,21 @@
...
+from fluidity_tools import stat_parser
+from pylab import show
+s = stat_parser("lagrangian_detectors.detectors")
+
+data = zeros((2,100,size(s['Steve_001']['position'][0])))
+
+for i in range(100):
+    n = i + 1
+    padding = ''
+    if(n<100):
```
cianwilson: Use ``'%03d'%n`` for the padding in one line.

tests/lagrangian_detectors_increasing_rotation/plotscript.py
```diff
@@ -0,0 +1,21 @@
...
+from fluidity_tools import stat_parser
+from pylab import show
+s = stat_parser("lagrangian_detectors.detectors")
+
+data = zeros((2,100,size(s['Steve_001']['position'][0])))
+
+for i in range(100):
+    n = i + 1
+    padding = ''
+    if(n<100):
```
cianwilson: Use ``'%03d'%n`` for the padding in one line.

femtools/Diagnostic_variables.F90
```diff
@@ -1643,7 +1643,8 @@ subroutine initialise_detectors(filename, state)
              if (default_stat%detector_group_names(j)==temp_name) then
                 read(default_stat%detector_checkpoint_unit) detector_location
                 call create_single_detector(default_stat%detector_list, xfield, &
-                      detector_location, i, STATIC_DETECTOR, trim(temp_name))                  
+                      detector_location, i, STATIC_DETECTOR, trim(temp_name))   
+                default_stat%detector_list%detector_names(i)=trim(temp_name)
```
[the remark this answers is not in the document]
Cmath2: The detector_list was given the names if initialized from an options file (under the logic if (.not.default_stat%from_checkpoint), this falls under the 'else' logic gate where the detectors are initialized from checkpoint. Previously detector_list%detector_names were only set for arrays of detectors (now line 1693 of the code), leading to issues where single detectors (whether stationary or lagrangian) were not correctly set up when reading from checkpoint. These two changes fix this bug, now assuring single detectors behave correctly when initializing from checkpoint.

detector_list%detector_names could potentially be set within the create_single_detector routine. It should be a fairly simple change if you'd like it done.

To your final point, I'm not sure if I completely follow. Detector_list is passed in to create_single_detector, and if performed in parallel, it returns if the picked element is not in the domain or not owned by the processor. If in serial, it checks if the element is within the domain, and if not, it checks if detector_list%write_nan_outside is enabled. If the element is not in the domain and write_nan_outside is disabled and FLExit is performed, whereas if the element is not in the domain and write_nan_outside is enabled, the detector is added to the list (with relevant parameters written as nan). To me this seems like exactly what we want to happen? Let me know if I've missed/misinterpreted something.

schemas/prescribed_field_options.rnc
```diff
@@ -85,6 +86,22 @@ prescribed_values_vector_field =
...
 
+prescribed_relaxation_term =
+   (
+      (
+         ## Temporal discretisation options
+      	 element temporal_discretisation {
+            ## Non-linear relaxation term
+            ##  = 0  -- previous timestep velocity solution used in non-linear terms of momentum equations
+            ##  = 1  -- previous iteration velocity solution used in non-linear terms of momentum equations
+            ##  0 <= ITHETA <= 1
```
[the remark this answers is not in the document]
Cmath2: Fixed below

tests/lagrangian_detectors_increasing_rotation/analytical_traj.py
Cmath2: Done below

Cmath2:
The goal of this pull request is two fold:
-1st is to fix the Runge-Kutta advection scheme for detectors which is currently not working as intended on master, followed by testing our updated scheme to ensure it works correctly
-2nd is to do a general cleanup of detector code and make some improvements/bug fixes.

----------------------------------------------------------------------------------------------------------------------
Part 1:
The Runge-Kutta advection scheme currently implemented in master only utilizes timestep-weights (the weights assigned to temporal positions in the Runge-Kutta method) and the Stage matrix itself (where weights are given to RK functions). This is missing a crucial component of the RK method, being the timestep-nodes. Timestep-nodes give the temporal positions at which the RK functions are applied, being somewhere between the current timestep (t) and the next timestep (t + dt). As an example, view the changes made to get_RK_traj.py, (which was also previously ignoring the timestep_nodes).

In master, RK functions are currently being calculated only using the current time (t). This gives inaccurate results when utilizing time-varying velocity fields, and ultimately leads to incorrect advection of detectors.

With the addition of timestep-nodes, these functions will instead be calculated at the correct temporal positions for a given RK advection scheme (e.g. for RK4 functions will be calculated at t, t+0.5*dt, t+0.5*dt, t+dt).

For more information here is a link to the wiki explaining each of the explicit RK parameters https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta_methods#Explicit_Runge%E2%80%93Kutta_methods

In order to perform these changes. an Old_Velocity field had to be implemented into Move_Lagrangian_Detectors.F90 (along with several other changes to initialize Old_Velocity and implement itheta throughout the code). With changes to the RK scheme implemented, several test cases were created to ensure the new RK scheme works as intended. This included two time-dependent velocity field advection tests (including an oscillating rotational velocity field and an increasing rotational velocity field). Both of these tests fail to accurately advect detectors when utilizing the old RK advection scheme, but once the RK scheme was updated both tests now pass.

Second and fourth order RK convergence tests were also added to test the convergence of our new RK scheme. The previous scheme contained only 1st order convergence, regardless the order of the RK scheme being utilized. The new scheme was found to converge at the correct order, being the order of the RK scheme utilized for advection (e.g. RK2 converges with second order, RK4 converges with fourth order).

----------------------------------------------------------------------------------------------------------------------
Part 2:
The second goal of this pull request is to do a general cleanup and some bug fixes of the detector code. This included:

-Editing of schema options to set detector binary output as default (and give the option to set ascii output)
-Shifting detector_list%detector_names to be set within create_single_detectors (this fixes a bug where single detectors would previously not have their names set if initialized from checkpoint
-Removal of several isparallel() statements within Diagnostic_variables.F90 (no longer needed with binary output change above)
-Removal of unnecessary output code relating to binary output in write_detectors (if output is binary, mpi_out is always used instead).

----------------------------------------------------------------------------------------------------------------------
Also as a side note, the test with adaptivity removed was one of the new time-dependent velocity advection tests (hence not showing up in the diff). Mentioning that it had adaptivity removed was obviously irrelevant to the PR.

Let me know if you'd like more information about anything or additional changes.

Best Regards,
Chris

cianwilson:
Should add that I'm assuming the buildbot's latest run is after these changes :-)

Cmath2:
Thanks Cian!

Yup latest buildbot is after the changes.

Best Regards,
Chris

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #180, opened 2018-03-07
Minimal changeset to configure and build successfully with python 3.x.

jrper:
I'm not sure whether I should really raise this as an issue first, but since I've just done the same for libspud, I thought it was worth a little effort to improve python3 compatibility in fluidity. This changeset does the absolute minimum needed to be able to configure and build successfully with PYTHON_VERSION=3.

Note that this python3 build actually leaves me with 18 semi-spurious test failures, due to code in the tests which is not python3 compatible (I *think* mostly due to mixed whitespace and tabs), but whether it's worth fixing them depends on whether this is functionality worth testing.


jrper:
To comment on my own pull request, the number of issues goes up rather further if `/usr/bin/env python` happens to point to python3 (which it shouldn't yet, but as I understand it, might one day. In that case there appear to be 255 `.xml`, `.flml`  and `.py` files in `tests` and `examples` which aren't valid python3, as well as issues with some of the tools, including `testharness` itself.

python/fluidity/state_types.py
```diff
@@ -20,8 +20,8 @@ def __init__(self,n=""):
   def __repr__(self):
     return '(State) %s' % self.name
   def print_fields(self):
-    print "scalar: ",self.scalar_fields,"\nvector:",self.vector_fields,"\ntensor:", \
-    self.tensor_fields,"\ncsr_matrices:", self.csr_matrices
+    print("scalar: ",self.scalar_fields,"\nvector:",self.vector_fields,"\ntensor:", \
```
stephankramer: This actually prints something else now in python2.7:
```
$ python -c 'print("scalar: ", 3, "\nvector:", 4)'
('scalar: ', 3, '\nvector:', 4)
$ python3 -c 'print("scalar: ", 3, "\nvector:", 4)'
scalar:  3 
vector: 4
```
so maybe add a `from __future__ import print_function` on top? Only an issue for print statements with more than one argument...

stephankramer:
Hm yeah, we'll have to bite the bullet on fixing these tests for python3 in the (not so far) future. It might be worth adding some linting on all .py files in the tree (we can discuss whether we want to go as far as PEP8). The python embedded in .xml and .flml is a little trickier unfortunately.

jrper:
I've actually got a Frankenstein script to do basic validation on .flml and .xml files now, which could be turned into a test, along the lines of the existing diamond_validation. Making it actually run flake8 might indeed be overkill though, not least for whoever ends up needing to fix everything.

jrper:
That was slightly painful, but this version now appears to pass the tests locally with both (python->python2.7, PYTHON_VERSION=2.7) and (python->python3, PYTHON_VERSION=3), at least when combined with the companion pull request for libspud and the libvtkfortran patch #170 . This does mean irrevocably breaking support for python2.6, but I hope we can live with that. @stephankramer, would you mind having another look? Also, does @tmbgreaves have any views on testing? This needs a buildbot queue anyway, but it would be nice to test in something with an ArchLinux style view of the universe.

tools/testharness.py
```diff
@@ -63,14 +64,14 @@ def __init__(self, length="any", parallel="any", exclude_tags=None,
...
-          print "Test criteria:"
-          print "-" * 80
-          print "length: ", length
-          print "parallel: ", parallel
-          print "tags to include: ", tags
-          print "tags to exclude: ", exclude_tags
-          print "-" * 80
-          print 
+          print("Test criteria:")
+          print(("-" * 80))
```
stephankramer: These seem to have ended up with double brackets

tests/python_validation/python_validate.py
```diff
@@ -0,0 +1,205 @@
...
+""" 
+
+python2 python_validate.py [-p] [-f] [file1 file2 file3]
+
+or 
+
+python3 python_validate.py [-p] [-f] [file1 file2 file3]
+
+-f  (--fixup)  Automatically attempt to fix obvious python3 errors using reindent and 2to3.
+-p  (--pep8)   Run flake8 to test adherence to pep8 coding stanaards.
```
stephankramer: typo

tests/python_validation/python_validate.py
```diff
@@ -0,0 +1,205 @@
...
+                return line
+        else:
+            return sep[0].join([sanitize(_, sep[1:]) for _ in line.split(sep[0])])
+
+    with open(filename, 'r') as temp:
+        for line in temp.readlines():
+            if len(line.strip())==0 or line.strip()[0]=='#':
+                pass
+            if 'print ' in line:
+                line = sanitize(line.rstrip('\n'), [';', '#'])+'\n'
```
stephankramer: Why is `#` treated as a separator? Surely if we ignore corner cases of `#` being inside a string (which is fair enough), then we might as well cut off at the `#` ? Also parsing for `#` should happen before `;`

tests/python_validation/python_validate.py
```diff
@@ -0,0 +1,205 @@
...
+        print(str(err))
+        fail = True
+        lines = snippet.split('\n')
+        nchar = len(str(len(lines)))
+        for k, line in enumerate(lines):
+            print(("%{0}d:".format(nchar))%(k+1), repr(line)[1:-1])
+    return fail
+
+def validate_file(filename, fix):
+    """Run the validotor on a whole file."""
```
stephankramer: spelling

tests/python_validation/python_validate.py
```diff
@@ -0,0 +1,205 @@
...
+        failed = process_snippet(filename, testfile.read())
+        if failed:
+            if fix:
+                print_as_function(filename)
+            return True
+    return False
+
+
+def validate_xml_snippets(filename, test_pep8, fix):
+    """Run the validotor on elements with attribute 'language=python' in an xml file."""
```
stephankramer: spelling

tests/python_validation/python_validate.py
```diff
@@ -0,0 +1,205 @@
...
+                failed =True
+                failures.append(filename)
+        else:
+            if validate_xml_snippets(filename, 
+                                     test_pep8=options.test_pep8,
+                                     fix=options.fix):
+                failed =True
+                failures.append(filename)
+
+    if failed:
```
stephankramer: `if failures` ? Non-empty lists evaluate as True

tests/python_validation/python_validate.py
```diff
@@ -0,0 +1,205 @@
...
+        print("Failed:")
+        for filename in failures:
+            print(filename)
+        exit(1)
+    else:
+        print("Passed!")
+        exit(0)
+
+def check_versions():
```
stephankramer: Is this actually used?

stephankramer:
Wow! That must have been quite a bit of work indeed. Looks very good. I'll add a buildbot branch to see if everything still works on python2.7. It would be a good idea to have this tested before/after merge with python3 indeed. Easiest way to do that (if we don't have python==python3 build environments) might be to simply create a venv with `python -m venv foo`, with that environment activated `python` should point to python3. Happy to break python2.6 indeed (if it currently works at all)

stephankramer:
http://buildbot.ese.ic.ac.uk:8080/builders/python3_compatibility

aclocal.m4
```diff
@@ -900,8 +899,8 @@ EOD`
 	AC_MSG_CHECKING(python extra linking flags)
 	if test -z "$PYTHON_EXTRA_LDFLAGS"; then
 		PYTHON_EXTRA_LDFLAGS=`$PYTHON -c "import distutils.sysconfig; \
-			conf = distutils.sysconfig.get_config_var; \
-			print (conf('LINKFORSHARED'))"`
+		conf = distutils.sysconfig.get_config_var; \
+		print ('' if conf('PYTHONFRAMEWORK') else conf('LINKFORSHARED'))"`
```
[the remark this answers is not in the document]
jrper: This is a somewhat selfish change that simplifies python configuration on Apple OSX, where python is installed as a "Framework" (read library plus metadata, installed in a non-POSIX location). It shouldn't affect anywhere which doesn't use framework installs.

tests/python_validation/python_validate.py
jrper: That certainly simplifies things.

tests/python_validation/python_validate.py
jrper: My intention was that check_versions be called as a warn_test to check python2 and python3 both exist on testing machines, with the validation then proceeding for all the versions which exist. The idea being that we can't expect people to validate a python version they don't have installed. The buildbot results show that isn't precisely happening as well as I'd hoped.

jrper:
I'll address the typographic comments shortly, as well as the diamond_validation failure on buildbot. Based on the results, it appears the instance the tests are run in doesn't include a version of python 3, which has both stress-tested my validator to the point of failure (I'll try and make it a bit more robust) and leaves some interesting questions about the optimum testing structure.

tmbgreaves:
I've added python3-dev to the packageset for the xenial testing environment, so once the dev-build job finishes it should be present in the branch tests.

jrper:
Thanks to @tmbgreaves  for the python3 change to buildbot, it's dealt with that particular snafu. @stephankramer I hope I've addressed everything you've raised in my latest commit.

 Regarding automated testing in a python3 universe, I'm hardly a docker expert, but it appears to be difficult to get sourcing the env activate script to stick. A small changeset I can get to perform is to add `python3-venv` to the packages in the the dev-build and then change the actual build to do

```
# Get the Fluidity source
RUN git clone $BRANCH https://github.com/$PROJECT/fluidity.git
WORKDIR /home/fluidity/fluidity
RUN git clone https://github.com/FluidityProject/longtests.git
RUN  python3 -m env python3
ENV VIRTUAL_ENV /home/fluidity/fluidity/python3
ENV PATH $VIRTUAL_ENV/bin:$PATH
RUN ./configure $CONFFLAGS
```

The real sticking point from here is that there aren't system packages for python3 VTK, which is only available with versions 7 and above. There *is* a pypi package, but it uses 8.1, which clashes with the system vtk6 libraries Fluidity gets built against. At the moment then, we'd have to build VTK ourselves.

P.S. @stephankramer any chance of having a look at the related upstream spud fix and at #170?

tmbgreaves:
Re. virtualenv activation in containers - setting the environment by hand is AFAIK the only way to do it, so what you're doing there looks right to me. If you want to be able to select at runtime you can ARG the relevant variables so they can be overridden, and then ENV them after the ARG line to make them 'stick' in the later runtime environment.

stephankramer:
Happy for this go. But I guess it would be a good idea to hold back any merges until we have some basic CI going again.

stephankramer:
@jrper are you happy for this to go in?

Re: testing, is there anything stopping us to go the whole hog, and replacing all `#!/usr/bin/env python` with `#!/usr/bin/env python3`. As far I understand the current changes aren't going to do much as system `python` will always be `python2` (I think it's only Arch being an outlier here, homebrew seems to have fixed their intransigence). Is python vtk still an issue?

jrper:
Fine by me for it to go in. Unfortunately there's still no sign of a system python3 vtk even for xenial, so if we switched to python3 right now I think we'd have to distribute our own.

Patol75:
What about [this one](https://packages.ubuntu.com/bionic/python3-vtk7) ?

PR #181, opened 2018-03-09
Petsc38 changes

stephankramer:
Changes needed to build fluidity with petsc v3.8.

The main difference in petsc's fortran interface  from v3.8 is the introduction of PETSC_NULL_XXX,
where XXX=KSP,Vec,Mat,etc. instead of the generic PETSC_NULL_OBJECT. Compatibility for older
petsc versions is maintained by #define-ing PETSC_NULL_XXX PETSC_NULL_OBJECT. Unfortunately a number of bugs/complications were discovered associated with this change that I needed to work around:
  - MatCreateSchurComplement is missing a hand-rolled fortran wrapper with a CHKFORTRANNULLOBJECT, so that a PETSC_NULL_MAT is not correctly translated to a C (Mat *) null
  - PetscObjectReference is also missing a wrapper (passing the %v member works instead)
  - there is not PETSC_NULL_MATNULLSPACE and MatGetNullSpace is not wrapped correctly, workaround with IsNullMatNullSpace() function that tests the returned nullspace
  - in petsc 3.8, the check that petsc object arguments are null arguments (and thus meant to be omited/ignored) is now checked by checking whether the value of the associated fortran derived type with single attribute %v, has the value -1. It is therefore possible that uninitialised petsc object variables in fortran are uninintenionally initialised to be null. This has unintended consequences for some petsc routines, such as MatCreateVecs, with optional output arguments. If the provided uninitialised argument happens to contain -1 as input value, it will be interpreted as null and therefore no output result is returned for it. Thus such arguments now need to be initialized with a fortran petsc object that is known not to be equal to PETSC_NULL_XXX
  - routines with missing wrappers to recognize PETSC_NULL_INTEGER: MatGetOwnerShipRange and MatSeqAIJSetPreallocation
  - no such thing as PETSC_NULL_RANDOM, so VecSetRandom *needs* a Random object as input
  - the (MPI_Comm *) comms argument of PCMGSetLevels does not have a clear null option in fortran, PETSC_NULL_KSP seems to work as there is actually no explcit wrapper

Most of these issues (in particular the uninitialised fortran arguments) have been addressed in https://bitbucket.org/petsc/petsc/pull-requests/829/added-f90-interfaces-for-several/diff These will not be available in this petsc release (3.8) however, as the solution to uninitialised fortran arguments (using derived type initialisation) requires disallowing petsc fortran objects in common blocks. So this merge sticks with the workarounds.

Other changes in this merge:
* remove petsc33 testharness tests as petsc <v3.4 is no longer supported and other small cleanups
associated with older petsc versions.
* removing some erronous cruft from configure
*  although a very simple test of petsc fortran modules was run in our configure (which prior to v3.8
decides whether the fortran modules are "functional", i.e. not compiled with a different fortran compiler)
the longer test was unintentionally not using modules. From petsc v3.8, petsc fortran modules are required and the --disable-petsc-fortran-modules fluidity configure option is no longer supported (icw petsc 3.8)
* removing previous workarounds to do with bugs in MatGetInfo and PETSC_NULL_OBJECT
* PCGAMGSetThreshold is changed (and PCGAMGSetThreshold is added), need to discuss possibly better default for Stokes. For now attempted to keep the same.

### DIFF
--- a/include/petsc_legacy.h
+++ b/include/petsc_legacy.h
@@ -6,13 +6,17 @@
 ! still need #ifdef PETSC_VERSION>... in the main code
 #include "petscversion.h"
 #ifdef HAVE_PETSC_MODULES
-#if PETSC_VERSION_MINOR>=6
+#if PETSC_VERSION_MINOR>=8
+#include "petsc/finclude/petsc.h"
+#elif PETSC_VERSION_MINOR>=6
 #include "petsc/finclude/petscdef.h"
 #else
 #include "finclude/petscdef.h"
 #endif
 #else
-#if PETSC_VERSION_MINOR>=6
+#if PETSC_VERSION_MINOR>=8
+#error "From PETSc v3.8, petsc fortran modules are required. Ensure petsc fortran modules, compiled with the same fortran compiler, are installed and configure without the --disable-petsc-fortran-modules option."
+#elif PETSC_VERSION_MINOR>=6
 #include "petsc/finclude/petsc.h"
 #else
 #include "finclude/petsc.h"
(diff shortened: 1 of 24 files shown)


stephankramer:
This is already build tested with petsc38 in the centos7-dev build (thanks @tmbgreaves !), here:

http://buildbot.ese.ic.ac.uk:8080/builders/centos7-standard

Build test with the current petsc 3.7 environment:

http://buildbot.ese.ic.ac.uk:8080/builders/petsc38-changes

Both are green

include/petsc_legacy.h
```diff
@@ -53,3 +57,20 @@
 #define PetscViewerAndFormatCreate NullPetscViewerAndFormatCreate
 #define PetscViewerAndFormatDestroy PETSC_NULL_FUNCTION
 #endif
+! from petsc 3.8 onward PETSC_NULL_OBJECT is gone, a specific PETSC_NULL_XXX needs to be used
+#if PETSC_VERSION_MINOR<8
```
jrper: Would it be better to include the major version in the test as well, just in case this is still around whenever petsc 4 happens?

adamcavendish: Major version testing included in #183 .

stephankramer:
Letting this run through buildbot once more with @adamcavendish 's changes. Any further comments? Otherwise could someone please approve when green?

jrper:
Looks fine to me now, and still green.

adamcavendish:
@jrper Is it possible to merge this branch back first?

stephankramer:
@adamcavendish this is merged now

stephankramer:
@adamcavendish sorry I see I accidentally merged an older version of the branch, missing out your commits. Have now merged these as well.

adamcavendish:
Great.

PR #183, opened 2018-03-22
Include PETSc major version in testing

adamcavendish:
PETSc major version is added in testing in case minor version testing only will break after 4.0 is released.


adamcavendish:
You're right. Updated.

PR #187, opened 2018-04-03
Add subtract_average and subtract_point_value diagn. algorithms

stephankramer:
These are useful when dealing with fields that are only defined up to a
constant (like pressure). During their solve it's usually better to use
the "remove_null_space" option, which adds a mesh-dependent constraint
that effectively means that the sum of the nodal values is zero. However
for diagnostic purposes it is often more useful to enforce a more well
defined constraint. This is now possible in two ways:

subtract_point_value - subtracts the value of the field value in a
specified (by coordinates) point, so that the resulting diagnostic field
is zero in that point. Similar to using a reference_node, but done
afterwards.

subtract_average - subtracts the integrated mean/average of the field,
so that the integral of the resulting diagnostic field is zero

Switched on in two different tests.


stephankramer:
http://buildbot.ese.ic.ac.uk:8080/waterfall?show=remean-diagnostics

Issue #174, opened 2018-01-25
Statplot graphical artifact


Patol75:
Dear all,

Most of you may be aware of a small graphical artifact that affects statplot. Currently, the displayed plot links the last point to the first one as displayed on the attached picture.
![oldstatplot](https://user-images.githubusercontent.com/23244688/35371134-75512fe2-01e6-11e8-8f36-030e89b9a2df.png)
A web search tended to point out that this behaviour is Gtk-related. I therefore tried to modify the python scripts (tools/statplot.py, python/fluidity/diagnostics/gui.py & python/fluidity/diagnostics/plotting.py) so that they could run with the latest version of Gtk.
I created a branch called Staplot_Gtk3 which contains the three modified scripts. Roughly, Gtk 3.0 is now used with GTK3Cairo backend in Matplotlib. I also used argparse instead of getopt as it is marked as deprecated by the new Python standard. The result is included below.
![newstatplot](https://user-images.githubusercontent.com/23244688/35371141-7c9ed146-01e6-11e8-80bc-c444c4793c0f.png)
There is however a problem: when I finished all the modifications, I had the issue that the statplot window would close without reopening whenever a field modification (i.e. x- or y-axis field) would be performed. I could not figure out how to fix it, but I found out that adding those two lines:
window.show_all(),
Gtk.main(), at the end of the _RefreshPlot def (StatplotWindow class, statplot.py) would help. It is however a mystery for me, as the first line in particular should have the whole program to crash because the variable window is not declared at this stage. Indeed, the program produces an error, but the statplot window stays opened and one can display each and every combination of variables he/she wants.

Would anyone have thoughts on this ? Thank you indeed for any.

Thomas

stephankramer:
Unfortunately I don't have much experience tinkering with gtk. Hopefully @cianwilson does? One request though: could you separate the changes you've made to these files into:
* changes needed to switch to gtk3
* changing from getopt to argparse
* arbitrary cleaning up and python style improvements.

at the very least into separate commits.

At the moment it's very hard to actually work out what you've changed. In general arbitrary cleaning of code is something that should be avoided in multi-developer projects such as this for a number of reasons:
* it makes it hard for other people to see what you've done
* when your commits are merged amongst many others it makes it hard to see what code has actually really changed over a certain time-window and who's done what (git annotate)
* chances of introducing unintended bugs and the work that inflicts on others typically doesn't outweigh the positives of having a nice looking, clean code

[...]

PR #190, opened 2018-04-23
Enhanced Statplot

Patol75:
Dear all,

Following #174, I propose here a new version of the Statplot graphical tool.
The previous version depended on two files (gui.py and plotting.py) that have now been removed (please mention if any other code depended on those two files). The new version is used exactly in the same fashion as the previous one:
```
statplot path/to/statfile
```
and has the following working properties:

- Ability to refresh the displayed data as the statfile is being modified by the Fluidity tasks (key 'r')

- Ability to close the active window (key 'q')

- Ability to switch between linear and logarithmic scales for both X and Y axes (keys 'x' and 'y'). Note however that switching to the logarithmic scale is only enabled for data ranges that have non-negative values only and that show at least an order of magnitude variation.

- Ability to switch between a line plot and a marker plot (keys 'l' and 'm')

I am happy to take any comment and to add or modify any feature.

Thomas


drhodrid:
Hi Thomas,

I tested this with a few of my stat files and, with your latest changes, everything seems to work well for me locally. I like it!

I would like to ask @jrper to take a look when he can, just to verify this works on a mac etc... 

Also, we should run this through buildbot prior to a merge. @tmbgreaves could you add a buildbot for this branch?

Patol75:
I have tried using Statplot on Ubuntu 18.04 Bionic, and apart from having to install two "extra" packages (```python-gi``` & ```python-gi-cairo```), it still runs fine when called with Python 2. It does not however run with Python 3 as few scripts in diagnostics have some conflicting syntax. Do you think it is worth modifying them so that Statplot can run with Python 3 ?

jrper:
There's no time like the present. Based on the discussions in #180 we'd definitely like to transition to full Python 3 support, and I can't remember how far I went with python3 proofing statplot in that PR. On the other hand, I'm fairly confident it does fix the basic syntax, so it's one place to start.

Patol75:
Alright, so I'll basically wait for your branch to be merged and then test statplot against the new files.

Patol75:
@tmbgreaves I have noticed the following message
```Required` context class hudson.FilePath is missing
Perhaps you forgot to surround the code with a step that provides this, such as: node```
at the bottom of the thread following the Jenkins upgrade. Does it mean something is missing inside the branch on my side ?

**Comment by tmbgreaves** [comment]
Having poked it, I *think* this was just a transient error as the re-run seems to be working OK. However, it has got me to spot that master branch is not behaving properly and junit isn't seeing the test output files for some reason ...

**Comment by jrper** [comment]
@tmbgreaves I suspect what you're seeing is a transient failure in the blue ocean UI. At least I can make it appear or disappear by refreshing the page. Certainly the classic view seems to record them ok.

**Comment by tmbgreaves** [comment]
@jrper that's a relief - not a good feeling spotting that one!

**Comment by drhodrid** [comment]
Just following up on this - @jrper - did your python 3 branch get merged into master? If not, could you provide a brief update on its status? As noted above, this particular merge is probably best done following the merge of your python 3 changes.

**Comment by jrper** [comment]
@drhodrid It's not merged yet due to issues with testing the python3 functionality on Jenkins. I'd say I'm working on it, but at this second the whole jenkins.ese.ic.ac.uk server doesn't seem to want to talk to me.

**Comment by Patol75** [comment]
Dear all,

I invite you to test and review the updated code which I have just pushed. I believe the branch is up to date with master, so there should not be much conflict while merging it.
Please let me know if anything is unclear in the code, if some features are not working properly, some others are missing, etc...

**Comment by drhodrid** [comment]
Looks good to me. Seems to work nicely - which cannot be said for the existing version. One important note: python-gi-cairo needs to be installed for this to work.

**Comment by drhodrid** [comment]
@jrper - would you mind having a play/look at this? I'm happy with it, but would be good to have another user (with a different system!) take a look?

File: tools/statplot.py [hunk 1]
```diff
@@ -1,229 +1,305 @@
...
+        self.PopulateCombo('load')
+        self.IniCompl(self.xCombo)
+        self.IniCompl(self.yCombo)
+        self.xCombo.connect('changed', self.ComboChangedX)
+        self.xCombo.connect('key-release-event', self.ReleaseFocus)
+        self.yCombo.connect('changed', self.ComboChangedY)
+        self.yCombo.connect('key-release-event', self.ReleaseFocus)
+        hbox.pack_start(self.xCombo, True, True, 0)
+        hbox.pack_start(self.yCombo, True, True, 0)
+        self.fig, self.ax = plt.subplots(nrows=1, ncols=1, num=0,
```
angus-g: This is something weird on one of my systems, but creating the figure this way gives me a segfault somewhere in the depths of gtk. Instead, something like the following works:
```suggestion
        from matplotlib.figure import Figure
        ...
        self.fig = Figure(figsize=(14,7))
        self.ax = self.fig.gca()
```

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
+        self.PlotData('create', self.PlotType, 'linear', 'linear')
+        self.toolbar = NavigationToolbar(self.canvas, self)
+        self.vbox.pack_start(self.toolbar, False, False, 0)
+
+    def ComboChangedX(self, widget):
+        complOptions = self.PopulateCompletion(self.xCombo.get_active_text())
+        dynModel = Gtk.ListStore.new([str])
+        for option in complOptions:
+            dynModel.append([option])
+        self.xCombo.get_child().get_completion().set_model(dynModel)
```
angus-g: I don't really understand this? You've already set up completion on the ComboBox, so adding a new Model on the Entry every time text is changed seems needless? Remove lines 77-81, and you still have perfectly functioning completion.

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
+            dynModel.append([option])
+        self.yCombo.get_child().get_completion().set_model(dynModel)
+        if self.yCombo.get_active_text() in sorted(self.entries.Paths()):
+            self.PlotData('update', self.PlotType, self.ax.get_xscale(),
+                          'linear')
+            self.canvas.grab_focus()
+
+    def FormatAxis(self, ax2fmt, scale):
+        if ax2fmt == 'x':
+            self.ax.set_scale = self.ax.set_xscale
```
angus-g: Recommend making this local, not modifying the `ax` object.

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
+        key = event.string
+        if key == 'r':
+            self.RefreshData(self.statfile)
+            self.PlotData('update', self.PlotType, self.ax.get_xscale(),
+                          self.ax.get_yscale())
+        elif key == 'q':
+            self.destroy()
+        elif key == 'x' or key == 'y':
+            if key == 'x':
+                self.ax.get_scale = self.ax.get_xscale
```
angus-g: Again, recommend this should be local

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
+        self.xCombo.connect('changed', self.ComboChangedX)
+        self.yCombo.connect('changed', self.ComboChangedY)
+
+    def ReleaseFocus(self, widget, event):
+        if event.keyval == Gdk.KEY_Escape:
+            self.canvas.grab_focus()
+
+
+window = StatplotWindow(args.statfile)
+window.connect('delete-event', Gtk.main_quit)
```
angus-g: Need to connect `destroy` here too to make the "q" keybinding work.

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
+                    failcount += 1
+            if failcount == 5:
+                raise Exception('Could not open %s' % filename)
+        if len(stats) == 1:
+            return stats[0]
+        else:
+            return fluidity_tools.JoinStat(*stats)
+
+    def RefreshData(self, statfile):
+        self.xCombo.disconnect_by_func(self.ComboChangedX)
```
angus-g: I think it's more elegant to just block the signal handlers:

```suggestion
        with self.xCombo.handler_block(self.handler_combochangedx), self.yCombo.handler_block(self.handler_combochangedy):
```

This way, the handler unblock is scoped automatically. This requires you to save the handler IDs in the constructor however (`Object.handler_block_by_func` doesn't define `__enter__`, etc.).

Alternatively. you could just block/unblock instead of disconnecting and reconnecting:

```suggestion
        self.xCombo.handler_block_by_func(self.ComboChangedX)
        ...
        self.xCombo.handler_unblock_by_func(self.ComboChangedX)
```

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
-            
-    return
-
-# The window
-window = StatplotWindow(args)
-window.set_default_size(640, 480)
-
-# Fire up the GUI
-gui.DisplayWindow(window)
+#!/usr/bin/env python2
```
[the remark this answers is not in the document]
Patol75: Do you mean leaving it as `#!/usr/bin/env python`? Ultimately I guess we want it to be `#!/usr/bin/env python3` to avoid any Python 2 execution?

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
+import numpy  # noqa: E402
+import sys  # noqa: E402
+import time  # noqa: E402
+import warnings  # noqa: E402
+
+rc('text', usetex=False)
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description='GUI for Fluidity .stat outputs\n\nWhen the window is '
```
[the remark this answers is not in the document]
Patol75: I had never realised this was possible, thank you.

tools/statplot.py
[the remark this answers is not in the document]
angus-g: Yes, this code is compatible with both Python 2 and Python 3 (ignoring the fluidity library not being this way), so I personally don't see the harm in leaving it as `#!/usr/bin/env python`.

tools/statplot.py
Patol75: Just out of curiosity, which versions of gi and matplotlib is this happening on?

angus-g: gi 3.32.1, matplotlib 3.1.1

I think using the OO-interface in this context is the right thing to do regardless.

tools/statplot.py
Patol75: Entirely right, thank you. Completion is actually doing everything by itself, I guess I wanted to help it, but as you mentioned, needless.

tools/statplot.py
Patol75: Sensible, thank you.

tools/statplot.py
```diff
@@ -1,229 +1,305 @@
...
+            failcount = 0
+            while failcount < 5:
+                try:
+                    stats.append(fluidity_tools.Stat(filename))
+                    break
+                except (TypeError, ValueError):
+                    time.sleep(0.2)
+                    failcount += 1
+            if failcount == 5:
+                raise Exception('Could not open %s' % filename)
```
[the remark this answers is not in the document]
Patol75: To be honest, I have not really paid attention to this function in the sense that I have never investigated why fluidity_tools.Stat would raise a TypeError or a ValueError. I will have a look and propose a re-write.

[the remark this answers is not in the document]
angus-g: I guess the general point here is that if the load fails too many times, we shouldn’t be masking the actual error, which makes it quite hard to debug

tools/statplot.py
Patol75: From this [documentation](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Widget.html#Gtk.Widget.signals.delete_event), I am not entirely sure on the difference between the two. But judging from [this](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Widget.html#Gtk.Widget.destroy), I guess I will just connect `destroy`? Also, do you mean `q` is not working for you? And a last one on this: if I close the window using Alt+F4, I get the below error message in the Terminal. Any idea?
`Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/matplotlib/backends/backend_gtk3agg.py", line 40, in on_draw_event
    self._render_figure(w, h)
  File "/usr/lib/python2.7/dist-packages/matplotlib/backends/backend_gtk3agg.py", line 31, in _render_figure
    backend_agg.FigureCanvasAgg.draw(self)
  File "/usr/lib/python2.7/dist-packages/matplotlib/backends/backend_agg.py", line 433, in draw
    toolbar.set_cursor(toolbar._lastCursor)
  File "/usr/lib/python2.7/dist-packages/matplotlib/backends/backend_gtk3.py", line 494, in set_cursor
    self.canvas.get_property("window").set_cursor(cursord[cursor])
AttributeError: 'NoneType' object has no attribute 'set_cursor'`

tools/statplot.py
Patol75: Would the following work?

Inside __init__
`self.xCon = self.xCombo.connect('changed', self.ComboChangedX)`
`self.yCon = self.yCombo.connect('changed', self.ComboChangedY)`

`def RefreshData(self, statfile):
        with (self.xCombo.handler_block(self.xCon),
              self.yCombo.handler_block(self.yCon)):
            self.xCombo.remove_all()
            self.yCombo.remove_all()
            self.entries = self.ReadData(statfile)
            self.PopulateCombo('reload',
                               prevIterX=self.xCombo.get_active_iter(),
                               prevIterY=self.yCombo.get_active_iter())`

angus-g: Yes, it should do. Glad you figured out my rather jumbled suggestion!

tools/statplot.py
angus-g: The `delete-event` is sent by the window manager (if you alt-F4, etc.). Pressing "q" is set up here to destroy the object, emitting the `destroy` signal instead. Indeed, for me, pressing "q" closed the window (because it was destroyde), but didn't quit the application. You should just need to connect `destroy` as well:

```suggestion
window.connect('delete-event', Gtk.main_quit)
window.connect('destroy', Gtk.main_quit)
```

I'm not really sure why you're getting those error messages, looks like the canvas isn't being cleaned up properly...

tools/statplot.py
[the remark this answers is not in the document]
Patol75: Would this work for you? The fifth try is simply out of the try/except statement.

```suggestion
                failcount = 0
            while failcount < 4:
                try:
                    stats.append(fluidity_tools.Stat(filename))
                    break
                except (TypeError, ValueError):
                    time.sleep(0.2)
                    failcount += 1
            stats.append(fluidity_tools.Stat(filename))
```

Patol75:
Thanks @angus-g for the review.
I have a suggestion here, do you think it is worth having the dependency on fluidity_tools for extracting information from the statfile? I have a starting point below that could be a substitution.

`from lxml import etree
from numpy import append, argsort, array, genfromtxt

entries = array([])
with open('statfile.stat', 'r') as fid:
    for num, line in enumerate(fid):
        if '<field' in line:
            info = etree.fromstring(line)
            if info.get('statistic') == 'value':
                entries = append(entries, info.get('name'))
            elif 'material_phase' in info.keys():
                entries = append(entries, info.get('material_phase') + '%'
                                 + info.get('name') + '%'
                                 + info.get('statistic'))
            else:
                entries = append(entries, info.get('name') + '%'
                                 + info.get('statistic'))
        elif '</header>' in line:
            break
values = genfromtxt('Pourquoi3D.stat', skip_header=num + 1)
values = values.T[argsort(entries)].T
entries = entries[argsort(entries)]`

Patol75:
As @angus-g approved, I will not commit anymore except if additional changes are required. However, I am attaching below a standalone version of statplot which has the ability to read through multiple similar statfiles on its own (removing the dependency onto fluidity.diagnostics.fluiditytools).
There are a few other fixes, such as denying the change to log scale below one order of magnitude instead of 2, adding 'a' and 'o' to grab focus on the ComboBoxes and setting completion.set_inline_selection to False. These last two changes allow navigating through the entries without having to use the mouse.
If you feel like any of these changes are relevant, I will be happy to commit them.

[statplot.txt](https://github.com/FluidityProject/fluidity/files/3459595/statplot.txt)

drhodrid:
Hi Thomas,

My advice here would be to get this merged, given things have been approved. If the additional changes that you mention would be useful to you and other users, they should be dealt with through another PR. For now, having a functioning version of statplot with gtk3 is a huge improvement.

Thank you once again for all of your efforts on this.

Best wishes,

Rhod

On 2 Aug 2019, at 5:16 am, Thomas Duvernay <notifications@github.com<mailto:notifications@github.com>> wrote:


As @angus-g<https://github.com/angus-g> approved, I will not commit anymore except if additional changes are required. However, I am attaching below a standalone version of statplot which has the ability to read through multiple similar statfiles on its own (removing the dependency onto fluidity.diagnostics.fluiditytools).
There are a few other fixes, such as denying the change to log scale below one order of magnitude instead of 2, adding 'a' and 'o' to grab focus on the ComboBoxes and setting completion.set_inline_selection to False. These last two changes allow navigating through the entries without having to use the mouse.
If you feel like any of these changes are relevant, I will be happy to commit them.

statplot.txt<https://github.com/FluidityProject/fluidity/files/3459595/statplot.txt>

Issue #119, opened 2016-05-09
libspatialindex creates temporary files when bulk loading


jrmaddison:
libspatialindex creates temporary files when bulk loading data beyond a certain size, which is quite inefficient. I'm not certain this is completely safe in parallel either, as two processes could potentially generate the same temporary filename (perhaps with very low probability) due to the use of mktemp:

```
./lib/libspatialindex.a(Tools.o): In function `Tools::TemporaryFile::TemporaryFile()':
Tools.cc:(.text+0x34ac): warning: the use of `mktemp' is dangerous, better use `mkstemp'
```

The following edit disables the use of disk caching by the libspatialindex bulk loader in a minimally invasive way:

```
index 50427ee..66de5fa 100644
--- a/spatialindex-1.8.0/src/rtree/RTree.cc
+++ b/spatialindex-1.8.0/src/rtree/RTree.cc
@@ -207,7 +207,7 @@ SpatialIndex::ISpatialIndex* SpatialIndex::RTree::createAndBulkLoadNewRTree(
        switch (m)
        {
        case BLM_STR:
-               bl.bulkLoadUsingSTR(static_cast<RTree*>(tree), stream, bindex, bleaf, 10000, 100);
+               bl.bulkLoadUsingSTR(static_cast<RTree*>(tree), stream, bindex, bleaf, std::numeric_limits<uint32_t>::max(), 1);
                break;
        default:
                throw Tools::IllegalArgumentException("createAndBulkLoadNewRTree: Unknown bulk load method.");
```

jrper:
Has this been fixed/changed upstream? I understand from @tmbgreaves that the version of spatialindex used in fluidity is getting fairly old.

jrmaddison:
This remains an issue when using libspatialindex 1.8.5.

jrmaddison:
Also this isn't a bug exactly - libspatialindex is designed to cache data on disk. The issue is that the cache size is not exposed in the bulk loader interface used by Fluidity.

jrper:
Is the alternative to fix fluidity to use the other overloaded method which takes a property set for input? Or does that do something we don't want?

jrmaddison:
I think you're right, that seems to be a way to do this.

PR #198, opened 2018-05-12
Build spring clean

jrper:
This changeset attempts to silence some of the superfluous warnings that fluidity currently spits out during building. The biggest things it doesn't deal with are an upstream warning from spud, and an issue caused by two different files in population_balance both producing a module with the same name. It also implements jrmaddison's suggestion in #119 to stop spatialindex from making temporary files.


jrper:
I've got no idea whether MPI still misbehaves enough anywhere to need to cd back to the working directory on any system we still support. If not, then I'd be *very* happy to remove it all. 

@tmbgreaves Any chance of a xenial buildbot queue for this, since it does have a non-zero chance of breaking something, and it would be nice to keep that green at least.

PR #203, opened 2018-06-01
Jenkins initialisation

tmbgreaves:
Adding Jenkins support for Fluidity. This commit adds a template Jenkinsfile which runs the basic tests on a non-debug build, working from base images on dockerhub built from Dockerfiles included in the docker/ directory in this repository. 

Issues with netcdf_read_errors are addressed, where the mount point for jenkins-derived filesystems in the docker build agent are long enough to trigger false warnings about broken /proc/mounts, which pollute the expected error output. A sed hack is introduced to deal with this. This problem *should* be dealt with by any recent hwloc as included in OpenMPI 3.0 and later, so the sed hack can probably be backed out in the not too distant future.

Three other tests looking at error output are returning wierd results and are provisionally turned off so as not to mask future failures, but the commit doing this should again probably be backed out once the issue is properly understood.

With the three missing tests and not yet covering the makefiles, debug, dpkg builds etc. this needs further work but to get some CI back I think it's worth getting this basic setup into master sooner rather than later.

As a side note - this finally adds Github integration for Fluidity CI, which I think is A Good Thing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,39 @@
+def PROJECT_NAME=JOB_NAME.split("\\s|/")[2].toLowerCase()
+
+pipeline {
+    agent { 
+        docker {
+            image "fluidity/baseimages:${PROJECT_NAME}"
+            label 'azure-linux'
+        } 
+    }
+    environment {
+        MPLBACKEND = 'PS'
+    }
+    stages {
+        stage('Configuring') {   
+            steps { 
+                sh './configure --enable-2d-adaptivity' 
+            }
+        }    
+        stage('Building') {       
+            steps { 
+                sh 'make -j' ;
+                sh 'make -j fltools' ;
+                sh 'make manual'
+            }
+        }
+        stage('Testing') {       
+            steps { 
+                sh 'make -j 8 unittest' ;
+                sh 'make -j 8 test' ;
+                sh 'make -j 8 mediumtest'
+            }
+        }
+    }
+    post {
+        always {
+            junit 'tests/test_result*xml'
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


jrper:
As a comment, I think imperfect but tested today is significantly better than perfectly tested eventually.

One query regarding desired failure modes. I think @stephankramer is ok with now with failed tests marking a branch as unstable. If that's not true, then it seems according to [this](https://support.cloudbees.com/hc/en-us/articles/218866667-How-to-abort-a-Pipeline-build-if-JUnit-tests-fail-) be possible to make things more binary by changing in the jenkinsfile post section to something like this (I've not tested this)

```

post {
         always {
            try {
               junit 'tests/test_result*xml'
           } catch(err) {
               step([$class: 'JUnitResultArchiver', testResults: 'tests/test_result*.xml'])
               if (currentBuild.result == 'UNSTABLE')
                   currentBuild.result = 'FAILURE'
               throw err
          }
       }
    }

File: Jenkinsfile [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+        stage('Building') {       
+            steps { 
+                sh 'make -j' ;
+                sh 'make -j fltools' ;
+                sh 'make manual'
+            }
+        }
+        stage('Testing') {       
+            steps { 
+                sh 'make -j 8 unittest' ;
```
stephankramer: I would prefer not to have the `-j 8` in there: I don't think it does anything (as make only sees a single target to be built) - so it's harmless, but not something we want to suggest others to use.

stephankramer:
I agree with @jrper . I think it's important to get these failures fixed (I've seen them as well locally) indeed, since missing error messages greatly hamper debugging - but I don't think it should hold up getting started with Jenkins. So let's just switch them off for the minute.

tmbgreaves:
Thanks both - I'll build the suggestions into the follow-up PR tweaking behaviour to work better for us.

PR #205, opened 2018-06-13
Test harness bug fix plus Jenkins slack integration

jrper:
These changes fix an important oversight in the testharness junit support, which meant two test failures on the Jenkins Centos7 build weren't being picked up. Further, it adds Slack notifications to the Jenkinsfile, as well as marking unstable builds as failed on Jenkins, in a method that could be extended if we want e.g email notifications of master build failures.

For the record, the 2 centos failures are `vtu2ensight`, due to missing files in the vtk support and `gls-ocean-param`, apparently due to `matplotlib` issues.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,6 +14,7 @@ pipeline {
     stages {
         stage('Configuring') {   
             steps { 
+                slackSend "Build started - ${env.JOB_NAME} ${env.BUILD_NUMBER} (<${env.BUILD_URL}|Open>)"
                 sh './configure --enable-2d-adaptivity' 
             }
         }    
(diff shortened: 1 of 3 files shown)


jrper:
@tmbgreaves In case it's any use from your end, I can make the `vtu2ensight` problem go away by switching to the EPEL red hat vtk packages rather than the fluidity ones.

jrper:
This is now showing as green on Jenkins and showing correctly red on CentOS. In fact it's also showing a couple of extra problems I'd missed there in `netcdf_read_free_surface.netcdf_read_free_surface and `prescribed_normal_flow_landslide`.

Thanks to @tmbgreaves for whatever you did to get the CI working again. Is there any chance of a review?

tmbgreaves:
This looks good, thanks James! Re. slack integration, not critical to this PR but I guess we need to do some setup on the Jenkins end in terms of slack access tokens.

PR #212, opened 2018-10-10
Fix #2

### LINKED ISSUE #2: Free surface mesh movement with LinearMomentum is broken
When running with a free surface with mesh movement, and using LinearMomentum with a reference density rho\=1.0 the contribution to the rhs of the continuity equation is not divided by rho in assemble/Free_Surface.F90, line 543. Without mesh movement in line 561, the term does get divided by rho. 

Note that with the variable_density option we already divide by rho in the mass matrix mass_ele, but that option does not work with mesh movement because of the logic in line 364. If there is some reason we don't support that combination we should probably options check for that.

Ultimately this of course indicates that fs+mesh movem.+linearmom. isn't really tested, so we need some validation+testing work for this to be functional.

stephankramer:
Still broken, but should now give a "not implemented error"

### DIFF
--- a/assemble/Free_Surface.F90
+++ b/assemble/Free_Surface.F90
@@ -368,8 +368,10 @@ contains
             use_fs_mesh=.false.
           end if
 
-          variable_density = have_option(trim(fs_option_path)//"/type[0]/variable_density") &
-                            .and. (.not.move_mesh)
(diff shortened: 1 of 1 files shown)


tmbgreaves:
Looks uncontroversial to me.

PR #218, opened 2018-10-17
PETSc>=3.9 changes

nkr1729:
MatSolverPackage is now MatSolverType
PCFactorSetMatSolverPackage is now PCFactorSetMatSolverType

Line 139 https://www.mcs.anl.gov/petsc/petsc-3.9/include/petscmat.h.html

### DIFF
--- a/femtools/Solvers.F90
+++ b/femtools/Solvers.F90
@@ -1860,7 +1860,11 @@ subroutine create_ksp_from_options(ksp, mat, pmat, solver_option_path, parallel,
     PC:: subpc
     MatNullSpace:: nullsp
     PCType:: pctype, hypretype
+#if (PETSC_VERSION_MAJOR<=3 && PETSC_VERSION_MINOR<=8)
     MatSolverPackage:: matsolverpackage
+#else
+    MatSolverType:: matsolvertype
+#endif
     PetscErrorCode:: ierr
     
     call get_option(trim(option_path)//'/name', pctype)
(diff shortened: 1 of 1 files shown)


nkralpy:
I just realised that 39eede3 was completely unnecessary.

stephankramer:
Hi there, thanks for contributing!

Fluidity's way of dealing with different petsc version's is a little peculiar. In principle we write to the petsc interface of the latest release (that we support!), and the inconsistencies in previous versions are then dealt with using macros. Have a look at include/petsc_legacy.h to get the idea. This may look a little hacky (and it is!), but the sprinkling of various `#if (PETSC_VERSION_MINOR=...` that we used to do ended up in an unmaintainable mess.

So the idea is that all fluidity source is currently compatible with petsc 3.8. If you want to upgrade our support to 3.9, just change the fortran source files directly (without #ifdefs) to what it should be under 3.9, and then in include/petsc_legacy.h add something like:
```
#if (PETSC_VERSION_MAJOR==3 && PETSC_VERSION_MINOR<9)
#define MatSolverType MatSolverPackage
#endif
```

By the looks of https://www.mcs.anl.gov/petsc/documentation/changes/39.html that seems to be the only change that affects fluidity. Unfortunately support for 3.10 might be a little more work.

Hope this makes sense. If you could try out the above, test it locally with your own petsc 3.9 build by running make shorttest, make mediumtest and make unittest, then once we have our own standard build up and running again we can merge it in master.

angus-g:
I am actually running PETSc 3.11, and other than this issue (and possibly #225) everything seems to work fine. I got a few test failures running on my laptop, but that might also be because I've only got 2 cores available and MPI complains on the oversubscribed tests...

Patol75:
@angus-g I am happy to give it a try on lots of cores if you point me toward the branch I should use.

stephankramer:
@nkr1729 thanks for flagging this up. This is now merged via #224. Let us know if you have any (more) issues with recent versions of petsc. We haven't yet added this to our test suite, so it's possible there are remaining issues - or failing tests with too strict tolerances due to answers slightly changing...

PR #221, opened 2019-03-20
Write extruded mesh only

stephankramer:
Two related changes:

* Add new write_extruded_mesh_only flredecomp option. This option will change the extruded mesh to the external from_file mesh and drop all horizontal meshes. It will still perform a columnar decomposition based on the horizontal input mesh. Rerunning flredecomp on the output flml of this first step however, it will decompose in all dimensions. This allows a two-stage flredecomp, where in the first stage we decompose to a enough cores that the decomposed extruded mesh can be fit in memory. In the second step we can then decompose to a larger number of cores that we actually want to run on. This must be done in two stages because in a single stage decomposition with an extruded setup, the final number of cores may be too large to decompose the horizontal mesh: not enough nodes per core, leading to empty partitions.
    
* This commit also fixes an issue with extrusion + arbitrary (i.e. not vertically_structured) adaptivity. The checkpoints written by this setup were using the same structure as a checkpoint with vertically structured adaptivity, or non-adaptive extruded runs: i.e. the extruded (and subsequently adapted mesh) is checkpointed and added under extrude/checkpoint_from_file in the options tree. The checkpoint flmls were functional for a restart (as tested in extrude_adapt3d_parallel/) but could not be flredecomped, as the decomposition of the horizontal and extruded mesh would be completely out of sync. To fix this, "arbitrary" adaptivity off an extruded initial mesh, now changes the options tree such that the extruded (and since  adapted) mesh becomes a from_file mesh (i.e. external mesh) and removes the horizontal meshes from the options tree. During the run the horizontal meshes are kept in state, as they are needed to complete a consistent .stat file. The .stat file from a pick up run however is necessarily different if the original (horizontal) InputMesh was included in stat in the initial run.

Test of the new option in tests/write_extruded_mesh_only. Test of the fixed extrude+adapt checkpoints in extrude_adapt3d_parallel


drhodrid:
@tmbgreaves any idea why we are not seeing a jenkins build for this?

tmbgreaves:
@drhodrid I don't know why, but Jenkins appears to scan the repository once and if a Jenkinsfile isn't found on a branch it doesn't appear to look at that branch until another manual scan happens. I re-ran the repository scan, and it's now picked up -- see if that's working as expected now?

Thanks for the heads-up!

drhodrid:
Hi Tim.  It now seems to be working as planned. Thank you.

R

PR #222, opened 2019-03-29
Fix normal orientation

stephankramer:
Fixes orientation of normals in the super-parametric case. In this case it is not necessarily true that the vector between element centroid and facet centroid is in the same direction as the normal. Instead choose orientation using the local coordinate associated with the opposite vertex (requires dn_s, derivatives of element shape functions evaluated at the facet).

### DIFF
--- a/tests/spherical_mappings_2d/spherical_mappings_2d.xml
+++ b/tests/spherical_mappings_2d/spherical_mappings_2d.xml
@@ -138,35 +138,38 @@ import numpy
 stat = stat_parser("super_analytic.stat")
 super_analytic_gradient_outer_surfaceintegrals = numpy.array([stat["State"]["RadiusP1"]["surface_integral%"+repr(i)][-1] for i in range(9,17)])
 </variable>
+<variable name="R1" language="python">R1=1.2</variable>
+<variable name="R2" language="python">R2=2.7</variable>
+<variable name="nsegments" language="python">nsegments=8</variable>
   </variables>
   <pass_tests>
     <test name="linear_integral" language="python">from math import pi
-print(abs(linear_integral-5.85*pi)/(5.85*pi))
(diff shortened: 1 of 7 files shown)


tests/spherical_mappings_2d/spherical_mappings_2d.xml
```diff
@@ -233,32 +236,32 @@ assert all(abs(analytic_gradient_inner_surfaceintegrals[1:]-analytic_gradient_in
...
-print(abs(super_gradient_inner_surfaceintegrals-0.3*pi)/(0.3*pi))
-assert all(abs(super_gradient_inner_surfaceintegrals-0.3*pi)/(0.3*pi) &lt; 0.07)
+print(abs(super_gradient_inner_surfaceintegrals-2*R1*pi/nsegments)/(2*R1*pi/nsegments))
+assert all(abs(super_gradient_inner_surfaceintegrals-2*R1*pi/nsegments)/(2*R1*pi/nsegments) &lt; 0.07)
 </test>
     <test name="super_gradient_outer_surfaceintegrals" language="python">from math import pi
-print(abs(super_gradient_outer_surfaceintegrals-0.675*pi)/(0.675*pi))
-assert all(abs(super_gradient_outer_surfaceintegrals-0.675*pi)/(0.675*pi) &lt; 0.003)
+print(abs(super_gradient_outer_surfaceintegrals-2*R2*pi/nsegments)/(2*R2*pi/nsegments))
+assert all(abs(super_gradient_outer_surfaceintegrals-2*R2*pi/nsegments)/(2*R2*pi/nsegments) &lt; 0.05)
```
cianwilson: Maybe idle curiosity but any idea why the error on this case alone increased?

stephankramer:
Rebased on master to see if it fixes Jenkins issue

PR #223, opened 2019-04-02
Update AC_PYTHON_DEVEL aclocal macro

angus-g:
Versions of Python 3 may be tagged according to their ABI version (PEP 3149), e.g. `3.7m`, however the current `AC_PYTHON_DEVEL` macro is unable to detect this, leading to a discrepancy between `PYTHON_VERSION` and the libraries it attempts to link against. This updates the macro to the latest aclocal-contrib version, while retaining the custom behaviour for OSX in 139814b57 ("Minimal changeset to configure and build successfully with python 3.x.", 2018-03-07).

I'm not really sure of the status of the autotools vs. CMake build systems, so hopefully this isn't misplaced!

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -603,12 +603,12 @@ CPPFLAGS=$tmpCPPFLAGS
 ])dnl ACX_hypre
 
 # ===========================================================================
-#         http://www.nongnu.org/autoconf-archive/ac_python_devel.html
+#     https://www.gnu.org/software/autoconf-archive/ax_python_devel.html
 # ===========================================================================
 #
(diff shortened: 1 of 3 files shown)


angus-g:
Just realised that I missed the change from `PYTHON_LDFLAGS` to `PYTHON_LIBS` amongst the other changes I'd made locally to get the build going

angus-g:
Should we merge this if we’re pushing ahead with Python 3 support?

stephankramer:
Yes, it was green indeed - I will merge now, but do you have the required permissions to do so as well?

PR #225, opened 2019-04-03
Correct types in call to PETSc (3.8+) VecScatterCreate()

angus-g:
I can't figure out why this compiles in its current state, although not on my configuration. Currently, a null `VecScatter` object is passed through to an `IS` parameter. This should affect PETSc versions 3.8 and up, where typed null objects were introduced.


angus-g:
I thought there might've been an interface change, but I can't readily see one. It's entirely possible that I'm just missing it... I actually haven't seen those implicit interface warnings either!

PR #228, opened 2019-04-11
Implementation of active particle scheme

Cmath2:
This pull request is currently posted to allow easier reviewer access to the code and comment on necessary changes. The current version of the branch is not ready for merging with master.

This pull request aims to implement an active particle scheme within the fluidity framework. The build extends upon the now merged passive particle scheme, and enables particles to interact with, and influence the values of fields during advection. This is done in three steps:
1. Relevant field values are interpolated onto particles at their current position in space.
2. Particles then calculate 'attributes', based on their position, the current simulation time, and interpolated field values.
3. Particle attributes are then interpolated back onto the nodes of a chosen mesh, with interpolated particle values able to generate new fields, or influence old ones.

With particles able to interact with and generate fields, limits must be placed on the number of particles present in set regions of the domain. This build has two primary components, the implementation of an active particle scheme able to interact with the fields of a simulation, and the implementation of spawning and deleting algorithms to control the number of particles present (per control volume). Currently, particles are spawned by cloning all parent particles that exist within a control volume and perturbing their location by a user specified amount. Conversely, particles are currently deleted by implementing a coin flip algorithm, where a coin is flipped for every particle within a control volume, and individual particles are deleted if the result is heads.

The schema has been updated to include new options for particle spawning and deleting (under each particle_group). Here the user can set which mesh particles will be counted on (currently only works with p1 meshes), as well as the minimum and maximum desired limits on the number of particles per control volume. The location of spawned particles is also able to be influenced, either by setting a radius around parent particles that new particles can spawn, or by allowing particles to spawn randomly within the current control volume. Caps can also be placed on which particle subgroups are spawned within a specific control volume if one subgroup is dominant (e.g., if a cap of 90% is set, if 90% (or above) of the particles within a specific control volume belong to one subgroup, only that subgroup will be spawned).

Options have also been added underneath diagnostic scalar fields, with two new algorithms, 'from_particles' and 'number_of_particles', now available. 
-The 'from_particles' algorithm implements the ratio method to interpolate the given particle_group and particle_attribute values onto a user specified mesh (Currently only works for p1 meshes). Two different interpolation options are also present here, being a weighted distance interpolation of particle attributes to nodes, or nearest neighbour interpolation. 
-The 'number_of_particles' algorithm simply counts the number of particles present per control volume, and sets this as the value for the relevant node. This enables users to track regions of a domain that are relatively dense, or sparse in particles.
An option has also been added to delete particles upon leaving the simulation domain instead of making them static.

write_diagnostics has also been updated to include the total number of particles present per particle_group, and for each particle_subgroup within the given particle_group. All relevant test files have also had their files updated to be consistent with the new schema.

The potential option to remove detectors entirely with the full implementation of particles is also up for discussion.

Summary of changes that have been made:
-Implementation of active particle scheme, particles attributes can now be interpolated onto a mesh and influence field values
-Implementation of particle spawning and deleting routines to manage the number of particles present in each control volume
-Updated schema file to include options for particle spawning/deleting and enable fields to be generated from particles
-Added test cases for utilizing active particles, as well as particle spawning/deleting
-Added the number of particles present in a simulation to the .stat file
-Updated test options files


drhodrid:
Closing this for now, pending a future and updated merge request from @Cmath2.

PR #231, opened 2019-04-23
Dockerfile ubuntu14

TensShinet:
I wrote a Dockerfile for ubuntu14.04. And it does work.

### DIFF
--- a/docker/Dockerfile.ubuntu14
+++ b/docker/Dockerfile.ubuntu14
@@ -1,0 +1,34 @@
+# DockerFile for a Fludity development container
+
+# Use a Trusty base image
+FROM ubuntu:trusty
+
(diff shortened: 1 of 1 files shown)


stephankramer:
Just out of curiosity: what are you using this for? I'm surprised fluidity still builds with petsc 3.4. I don't think we actually want to support this I'm afraid. In particular with further python3 dependencies coming up.

TensShinet:
Nothing.Just a Suppercomputer Challenge in Asia. The organizer of this Challenge recommanded us to use fluidity in ubuntu14.04. So I wrote a Dockerfile to build this environment.

stephankramer:
I suspect this no longer works now that we depend on python3.

Issue #233, opened 2019-05-16
Explicit GMSH file format for tests?


angus-g:
While Fluidity requires msh version 2 formatted files as input, some versions of GMSH will output a newer (and thus incompatible) format by default. However, GMSH supports the `-format msh2` flag and option to explicitly generate the correct version.

I'm not sure how this interacts with the versions of GMSH people currently use -- it seems like 2.x is the norm, although the upstream project is up to 4.x (and indeed that's the one I've had pulled in on my local machine).

This change really boils down to running `sed -ie 's/\bgmsh\b/gmsh -format msh2/' */Makefile` in the project root, but I wanted to check if this is sensible first before submitting a rather ugly pull request that touches every test...

jhill1:
Would it be better/feasible to write a gmsh reader for the newer format? It would take longer (and I'm not volunteering to write it!) but it may be preferable to editing every test and make it easier to users in the long run.

stephankramer:
Agree, it would probably be good to look into that and depending on how different the format is to support one or both. Note that we need both reading *and* writing (for checkpoints and flredecomp).

jrper:
Essentially the biggest difference is a switch to using a more compressed, entity based system for tags, even in the ASCII format. It would be a finite, but non-trivial exercise to make a reader for both 2.X and 4.X formats.

In principle, one option would be to switch to using libgmsh to handle the actual file reading, now that the API is finished, but that would add a nasty dependency.

jrper:
See c41e5e54fc7e169c620b10d30058d7fd2918f788 for an attempt at a reader for GMSH file formats 4 and 4.1  (i.e. the new ones). Having played with libgmsh a bit, I suspect that a hand rolled reader is the better way to go, but testing it is going to be a nightmare (I'm confident that it's trivial to break this implementation).

angus-g:
Nice! As long as the input files don't get too esoteric I agree that this is probably the better approach. I'll have a bit of a play and see whether I can break it.

I could replicate the functionality with libgmsh pretty easily (although I didn't bother setting it up as a dependency, which seems much more painful). In this case, libgmsh is pretty simple to use (but very overkill) for this case: open a file and read off the nodes/elements. Memory management without the C++ API is a bit more painful since they don't actually detail how it works in the C API (I assume you're allowed to just free everything they allocate for you))...

Patol75:
Just a bit of bookkeeping here, but should this issue be closed given #235?

PR #235, opened 2019-07-05
GMSH v4 support

angus-g:
This builds on the issue raised in #233 and https://github.com/FluidityProject/fluidity/commit/c41e5e54fc7e169c620b10d30058d7fd2918f788. On my local build, all tests pass except for one on `spherical_segment`, due to some changes in GMSH itself.

This support would introduce a new dependency on the Python [meshio](https://github.com/nschloe/meshio) library for the `gmsh_mesh_transform` and `transform_mesh` tools, which currently expect ASCII MSH files and modify them directly. I think it's probably a more robust solution to abstract these mesh modifications through some kind of library anyway.

### DIFF
--- a/femtools/Read_GMSH.F90
+++ b/femtools/Read_GMSH.F90
@@ -840,8 +840,6 @@ subroutine read_nodes_coords_v4_binary( fd, filename, versionNumber, nodes )
     integer :: entityDim, entityTag, tag_index
     integer :: numentityelements
 
-    integer, allocatable :: ltags(:)
-
     read(fd,*) charBuf
     if( trim(charBuf)/="$Elements" ) then
        FLExit("Error: cannot find '$Elements' in GMSH mesh file")
@@ -877,11 +875,6 @@ subroutine read_nodes_coords_v4_binary( fd, filename, versionNumber, nodes )
        tag_index = fetch(entityMap(entityDim+1), entityTag)
        numTags = entityTags(tag_index)
 
-       allocate(ltags(numTags))
-       do i=1, numTags
-          ltags(i) = entityTags(tag_index+i)
-       end do
-
        do k=1, numEntityElements
           e = e + 1
           ! Read in whole line into a string buffer
@@ -896,12 +889,9 @@ subroutine read_nodes_coords_v4_binary( fd, filename, versionNumber, nodes )
           
           ! Now read in tags and node IDs
           read(charBuf, *) allElements(e)%elementID, allElements(e)%nodeIDs
-          allElements(e)%tags(:) = ltags
+          allElements(e)%tags = entityTags(tag_index+1:tag_index+numTags)
 
        end do
-
-       deallocate(ltags)
-
     end do
 
     ! Check for $EndElements tag
@@ -939,7 +929,6 @@ subroutine read_nodes_coords_v4_binary( fd, filename, versionNumber, nodes )
     integer :: entityDim, entityTag, tag_index
     integer(kind=c_long) ::  ltmp, numentityelements
 
-    integer, allocatable :: ltags(:)
     integer(kind=c_long), allocatable :: vltmp(:)
 
     read(fd,*) charBuf
--- a/femtools/Read_GMSH.F90
+++ b/femtools/Read_GMSH.F90
@@ -31,9 +31,12 @@ module read_gmsh
   ! This module reads GMSH files and results in a vector field of
   ! positions.
 
+  use iso_c_binding
   use fldebug
   use global_parameters, only : OPTION_PATH_LEN
   use futils
+  use data_structures
+  use linked_lists
   use quadrature
   use elements
   use spud
@@ -53,6 +56,12 @@ module read_gmsh
   public :: read_gmsh_file
 
   integer, parameter:: GMSH_LINE=1, GMSH_TRIANGLE=2, GMSH_QUAD=3, GMSH_TET=4, GMSH_HEX=5, GMSH_NODE=15
+
+  type version
+
+     integer :: major, minor
+
+  end type version
 
 contains
 
@@ -90,7 +99,11 @@ contains
     logical :: haveBounds, haveElementOwners, haveRegionIDs
     integer :: dim, coordinate_dim, gdim
     integer :: gmshFormat
+    type(version) :: versionNumber
     integer :: n, d, e, f, nodeID
+
+    type(integer_hash_table) :: entityMap(4)
+    integer, allocatable  :: entityTags(:)
 
     type(GMSHnode), pointer :: nodes(:)
     type(GMSHelement), pointer :: elements(:), faces(:)
@@ -111,20 +124,55 @@ contains
          access="stream", form="formatted" )
 
     ! Read in header information, and validate
-    call read_header( fd, lfilename, gmshFormat )
+    call read_header( fd, lfilename, gmshFormat, versionNumber )
+
+    if (versionNumber%major .eq. 4) then
+       do n=1,4
+          call allocate(entityMap(n))
+       end do
+       call read_entities( fd, lfilename, gmshFormat, versionNumber, &
+            entityMap, entityTags )
+    end if
 
     ! Read in the nodes
-    call read_nodes_coords( fd, lfilename, gmshFormat, nodes )
+    if (versionNumber%major .eq. 4) then
+       if( gmshFormat == asciiFormat ) then
+          call read_nodes_coords_version_four_ascii( fd, lfilename, &
+               versionNumber, nodes )
+       else
+          call read_nodes_coords_version_four_binary( fd, lfilename, &
+               versionNumber, nodes )
+       end if
+    else
+       call read_nodes_coords_version_two( fd, lfilename, gmshFormat, nodes )
+    end if
 
     ! Read in elements
-    call read_faces_and_elements( fd, lfilename, gmshFormat, &
+    if (versionNumber%major .eq. 4) then
+       if( gmshFormat == asciiFormat ) then
+          call read_faces_and_elements_v4_ascii( fd, lfilename, &
+               versionNumber, elements, faces, dim, entityMap, entityTags)
+       else
+          call read_faces_and_elements_v4_binary( fd, lfilename, &
+               versionNumber, elements, faces, dim, entityMap, entityTags)
+       end if
+    else
+       call read_faces_and_elements_version_two( fd, lfilename, gmshFormat, &
          elements, faces, dim)
+    end if
 
     call read_node_column_IDs( fd, lfilename, gmshFormat, nodes )
 
     ! According to fluidity/bin/gmsh2triangle, Fluidity doesn't need
     ! anything past $EndElements, so we close the file.
     close( fd )
+
+    if (versionNumber%major .eq. 4) then
+       do n=1,4
+          call deallocate(entityMap(n))
+       end do
+       deallocate(entityTags)
+    end if
 
 
     numNodes = size(nodes)
@@ -290,14 +338,14 @@ contains
   ! Read through the head to decide whether binary or ASCII, and decide
   ! whether this looks like a GMSH mesh file or not.
 
-  subroutine read_header( fd, lfilename, gmshFormat )
+  subroutine read_header( fd, lfilename, gmshFormat, versionNumber )
     integer fd, gmshFormat
 
     character(len=*) :: lfilename
     character(len=longStringLen) :: charBuf
     character :: newlineChar
-    integer gmshFileType, gmshDataSize, one
-    real versionNumber
+    integer gmshFileType, gmshDataSize, one, i
+    type(version) versionNumber
 
 
     ! Error checking ...
@@ -309,9 +357,18 @@ contains
 
     read(fd, *) charBuf, gmshFileType, gmshDataSize
 
-    read(charBuf,*) versionNumber
-    if( versionNumber .lt. 2.0 .or. versionNumber .ge. 3.0 ) then
-       FLExit("Error: GMSH mesh version must be 2.x")
+    do i=1, len_trim(charbuf)
+       if (charbuf(i:i) == '.') charbuf(i:i) = ' '
+    end do
+
+    read(charBuf,*, pad='yes') versionNumber%major, versionNumber%minor
+    if( versionNumber%major .lt. 2 .or. &
+         versionNumber%major == 3 .or. &
+         (versionNumber%major == 4 .and. versionNumber%minor .gt. 1) .or. &
+         versionNumber%major .gt. 4 &
+       ) then
+         
+       FLExit("Error: GMSH mesh version must be 2.x or 4.x")
     end if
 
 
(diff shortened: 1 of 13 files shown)


angus-g:
Looks like https://github.com/nschloe/meshio/pull/400 may remove the explicit requirement for the msh2 format in `shelf_meshtransform_freesurfaceinitialcondition`

femtools/Read_GMSH.F90
```diff
@@ -343,17 +410,139 @@ subroutine read_header( fd, lfilename, gmshFormat )
...
+
+  subroutine read_entities( fd, filename, gmshFormat, versionNumber, &
+       entityMap, entityTags )
+
+    integer, intent(in) :: fd, gmshFormat
+    type(version), intent(in)    :: versionNumber
+    character(len=*), intent(in) :: filename
+
+    type(integer_hash_table) :: entityMap(4)
+    integer, allocatable :: entityTags(:)
```
stephankramer: `intent(out)` I believe? Same for `entityMap`

femtools/Read_GMSH.F90
```diff
@@ -290,14 +339,15 @@ end function read_gmsh_simple
...
+  subroutine read_header( fd, lfilename, gmshFormat, versionNumber )
     integer fd, gmshFormat
 
     character(len=*) :: lfilename
     character(len=longStringLen) :: charBuf
     character :: newlineChar
-    integer gmshFileType, gmshDataSize, one
-    real versionNumber
+    integer :: gmshFileType, gmshDataSize, one, i
+    logical :: decimalVersion = .false.
```
stephankramer: This is dangerous (a bit of a pitfall in fortran): when using an initialisation statement in the declaration the variable automatically acquires the save attribute. This means that decimalversion is only set to .false. *once* and retains whatever it was set to in the previous call. Reading the code below, I _think_ what will happen is that if we read a "decimalversion" file first, the next file that is read it will a decimalversion regardless of whether there is a dot.

femtools/Read_GMSH.F90
```diff
@@ -111,21 +125,56 @@ function read_gmsh_simple( filename, quad_degree, &
          access="stream", form="formatted" )
 
     ! Read in header information, and validate
-    call read_header( fd, lfilename, gmshFormat )
+    call read_header( fd, lfilename, gmshFormat, versionNumber )
+
+    if (versionNumber%major .eq. 4) then
```
stephankramer: I know this isn't adhered to in the rest of this module (which is generally a bit messy), but can we stick with ==, /=, < and > rather than .eq., .neq., .lt. and .gt.

femtools/Read_GMSH.F90
stephankramer: Again, something not adhered to in the rest of the module. But I think as a minimum code style guide for subroutines we try to use:
- there should be at least a single comment line on top telling what the subroutine does (unless completely obvious)
- all arguments should have intent declared - except pointer argument for which declaring intent has only been recently allowed in the fortran standard (and refers to whether the association status is changed, not changes into the thing that it points to)
- all arguments should be declared on top, before any other local variables

stephankramer:
On the meshio dependency: it looks like bionic (and Debian buster) only have a python3-meshio package, not python-meshio. I'm actually in favour of breaking with python2 support, but then we first need to properly enforce it by changing `#!/usr/bin/env python` in `#!/usr/bin/env python3` everywhere...

Patol75:
Just a comment here, but you probably know about it, Python 2 support will be entirely dropped on Jan 1st, 2020 (https://www.python.org/dev/peps/pep-0373/). So it is indeed a good idea to complete the move to Python3.

femtools/Read_GMSH.F90
angus-g: Wow, I wasn't aware of that!

femtools/Read_GMSH.F90
angus-g: No problem, I can change these.

femtools/Read_GMSH.F90
angus-g: I'll claim some innocence by saying I didn't write most of the code! These are, however, good conventions and I'll make sure this PR is following them.

Patol75: On this topic, I found this [webpage](http://www.cs.rpi.edu/~szymansk/OOF90/bugs.html) which includes what @stephankramer explained and lists another few similar issues. I thought it was relevant to share. And indeed thanks to @stephankramer for the knowledge. :)

stephankramer:
I guess ideally the python3 she-bang stuff would have been in a separate merge - is that what you mean by "I misinterpreted your comment"? - but I'm perfectly happy for it to go in, in one go.

I think I've fixed the python_diagnostic test. This was testing whether we can read the options tree inside fluidity's embedded python. I was barking up the wrong tree for a while, as I thought it had something to do with static vs. dynamic linking of libspud and "import libspud" talking to a different dynamically loaded library than the one used by fluidity. It turns out that doesn't matter, as there is actually a mechanism to pass the handle of fluidity's options tree, via a "spud_manager" python module which is created on the fly inside fluidity's embedded python environment. Unfortunately this turned out to be a bit trickier to achieve in python3.

Could you resolve the conflicts and see if all tests pass now?

angus-g:
I do agree that the Python 3 stuff should be separate. I'm happy to pluck off the last two commits and put them in separately, then we can merge this one after that? I will however fix the conflicts and make sure the tests pass here anyway.

angus-g:
Well, it looks like Jenkins isn't running tests any longer? I can confirm however that your fix to the `python_diagnostic` test works, @stephankramer. All the short tests bar one pass with GMSH 4, which is down to a tolerance issue in spherical patch:

```
spherical_patch: Running fs_integral_run12:
Asserting relative error is smaller than 2e-05
relative error: 2.137004156780418e-05
spherical_patch: failure.
```

I did already make note of this in 904b2da58e5129ee7ce7e9cb13a9bab1aaae1684, which was the geometry change required to make that test run in the first place.

So, should the way forward here be for me to first slice off 5f867c6 and a669470 into their own PR? I think there may be a couple more changes required: I noticed that the `change_options` script in the `square-convection` test isn't Python 3-compatible. Was there anything outstanding GMSH-related in this PR?

angus-g:
Okay, I have split off the Python 3 specifics to a #238.

I think there is a test requiring the `gmsh2triangle` script which doesn't handle GMSH 4 files that I still need to fix for this present PR.

angus-g:
I've had a little look at updating `gmsh2triangle`. meshio does support writing out a mesh to the TetGen format, but it refuses to deal with meshes that don't actually contain any tetrahedra. It does seem like `gmsh2triangle` writes non-standard output files, putting in whatever element type came from the input file in the output `.ele` file. I did notice that it was marked deprecated in the manual, so I will let it be: msh2-formatted files will have to be a requirement for it.

drhodrid:
Hi Angus,

I think the best thing here would be to update the meshconv test (I assume that’s the one giving your grief?) to remove the gmsh2triangle components. Several years ago, Fluidity used to read in triangle files (as well as Gmsh), but that dependency was dropped once Gmsh became the mesh generator of choice for most (if not all) users. There is no obvious motivation to keep that aspect of the meshconv test in the repo.

Best wishes,

Rhod


On 23 Sep 2019, at 3:38 pm, Angus Gibson <notifications@github.com<mailto:notifications@github.com>> wrote:


I've had a little look at updating gmsh2triangle. meshio does support writing out a mesh to the TetGen format, but it refuses to deal with meshes that don't actually contain any tetrahedra. It does seem like gmsh2triangle writes non-standard output files, putting in whatever element type came from the input file in the output .ele file. I did notice that it was marked deprecated in the manual, so I will let it be: msh2-formatted files will have to be a requirement for it.

angus-g:
Thanks @drhodrid, that makes sense. There are quite a few tests that (I guess needlessly?) call `gmsh2triangle` in their Makefiles. Those are generating errors in the Makefiles, but the tests do seem to be passing regardless. You are right that the meshconv test seems to be the only one with an actual dependency on that script, and it is indeed actually failing. I'll remove references to `gmsh2triangle` from the following:

```
tests/channel-flow-dg/channel.py
tests/channel-flow-dg/channel_viscous.py
tests/diffusion-dg-stretched/makebox.py
tests/mms_tracer_P1dg_cdg_diff_steady_3d_cjc/cdg3d.py
tests/mms_tracer_P1dg_cdg_diff_steady_3d_cjc_inhNmnnbc/cdg3d.py
```

drhodrid:
@tmbgreaves any chance you could look into the jenkins failure here too? All tests seem to be green, but something goes funky yellow in the final "testing" stage. There doesn't seem to be any verbose output (unless I'm looking in the wrong place - likely!). Any ideas?

tmbgreaves:
Wierd - it's as if the junit output has vanished. I wonder if I broke it somehow with the latest round of base container updates? I'll hack at it.

tmbgreaves:
I'm seeing meshio packaged for python3 (python3-meshio) but not for python2 - is that expected?

tmbgreaves:
Possibly as an extension to this, though might be out of the scope of this PR (@stephankramer, @jrper ?) do we want to switch testing to bionic (18.04) to get default python 3?

drhodrid:
@tmbgreaves my feeling is that yes, we should switch all of our testing over to python3. Support for python2 will soon be dropped (https://pythonclock.org/). Many of Angus' merge requests rely on python 3 (gmsh relies on meshio, which is not packaged for python 2), and the particle stuff that we've done is largely tested in a python3 environment. So yes, unless there are any objections, let's make this switch.

drhodrid:
> I'm seeing meshio packaged for python3 (python3-meshio) but not for python2 - is that expected?

See previous response - yes, this is expected.

tmbgreaves:
@drhodrid OK! I've opened up a PR into angus-g:GMSH_V4_reader with the bionic switch.

tmbgreaves:
In the process of building with python3-meshio included. A sweep through the test failures *looks* like this is the only missing package now.

tmbgreaves:
@stephankramer @jrper comments solicited on whether this is a good moment to flip testing over to bionic

tmbgreaves:
> See previous response - yes, this is expected.

Whoops, yes, that was indeed the point of this whole discussion :-)

tmbgreaves:
Another update having worked through the packaging process -- python3-meshio is only available for bionic, not xenial, so we do need to link this PR with the bump to bionic. I'm in the process of rebuilding fluidity-dev for bionic at the moment and will push when complete.

drhodrid:
@tmbgreaves it looks like we've not lost the Jenkins link above altogether... how are things looking with this?

tmbgreaves:
Something very odd is happening there, I think on the Jenkins side. Having had a few cursory glances at it during the day I'm attempting some proper debugging now.

tmbgreaves:
Jenkins fixed and looking healthy so far -- I'll check again once the run is complete.

angus-g:
Looks like it gained 3 failures, but also still shows the same error on the Jenkins side?

tmbgreaves:
Some of those are admin error on my side, sorry -- I'd rebuilt the bionic base image but failed to push the latest version. That's on the way to dockerhub now.

tmbgreaves:
There look to be a couple of (hopefully minor) test failures resulting from the move over to bionic.

One is a newer test - @Cmath2 I'm seeing a small change in particle_attributes_zoltan where sq_error_4 is increasing from 0.000197026074023 (on xenial) to 0.00554233563029 (on bionic). Is it reasonable here to relax the tolerances?

The other is an old test from Jon Hill which I suspect is now unmaintained, but I'll drop Jon an email to see if he's still interested. I think it should not prevent merging this PR, though, as it's entirely unrelated to the main PR content.

tmbgreaves:
Summary of the remaining failures:

- diamond-validation is my problem with a missing system package; fixing now
- the gls medium test needs someone to get eyes on it and see if it just converged early to a physically reasonable solution -- I'll try and take a look.
- particle_attributes_zoltan needs @Cmath2 to take a quick look and see if we can relax tolerances
- the remaining three are all hitting a meshio import failure in transform_mesh which I *think* is a result of it being called with python rather than python3 from /usr/bin/env. I tried a fix but then hit a numpy error -- @angus-g , could you have a quick look at that one please? I suspect gmsh_mesh_transform will I think hit the same problem (and presumably isn't actually tested since we haven't hit the problem there..).

Unless we get issues resulting from those fixes I think that's the remaining task list for this PR.

Cmath2:
Hi @tmbgreaves @angus-g. I had a look at the test and can recreate the error. I've adjusted the tolerance for the failing test to 6.0e-3 (which should now pass).

angus-g:
Looks like using meshio to write the new mesh in those tools means that the new file doesn't have the `Entities` section, that `read_gmsh` is expecting to see (although it is technically optionally per the msh specification...)

I'll fix this up

tmbgreaves:
Update from Jon HIll via email -- he's going to take a look at the gls test, but likely not for a week or so. Since this is entirely independent of this PR I'd suggest that if you're ready to merge before Jon has pushed a fix, and the gls test is the only remaining failure, we land this PR and assume master is existing-broken so turning it red (with the bionic switch) is acceptable in the short term.

drhodrid:
Thanks for following up Tim. I think the preferred strategy is to get the python3 compat branch in first (given that this gmsh pr depends on python3-meshio), followed by this PR… there currently seems to be a build failure on the python3-comapt branch though (python version < 3.6)… so we need to figure out which version we want to require there first… unless I’m missing something…

On 11 Oct 2019, at 6:42 pm, Tim Greaves <notifications@github.com<mailto:notifications@github.com>> wrote:


Update from Jon HIll via email -- he's going to take a look at the gls test, but likely not for a week or so. Since this is entirely independent of this PR I'd suggest that if you're ready to merge before Jon has pushed a fix, and the gls test is the only remaining failure, we land this PR and assume master is existing-broken so turning it red (with the bionic switch) is acceptable in the short term.

tmbgreaves:
@Cmath2 confirming your fix is good in Jenkins -- thank you!

The existing diamond_validation failure is fixed but there's a new one which I'll do my best to address today.

angus-g:
Well, this is proving to be a little trickier than I'd hoped! The file is opened in formatted stream IO, which should let us read from a position we've saved previously:

```f90
inquire(fd, pos=before_header)
read(fd, *) charBuf
if (trim(charBuf) == "$Entities")
...

read (fd, *, pos=before_header) charBuf
```

However, the presence of mixed binary/ascii in the binary MSH format confuses gfortran, which probably uses those binary characters as extra record delimiters. In this case, the two (trimmed) reads are `$Nodes` (which we'd expect), and `yy$Nodes` when specifying `pos`. I'm not sure if this is a gfortran bug or not, but I couldn't reproduce it with ifort on a minimal working example.

drhodrid:
From memory, with a rebase, this one should also be good to go (pending green jenkins).

angus-g:
Sure, I'll rebase this now

angus-g:
Not quite green, we have a failure in one of the medium tests:

```
make[1]: Entering directory '/home/jenkins/workspace/fluidity_PR-235/tests/wind_stratification'
...
../../bin/transform_mesh '(x,0.001*y,z)' square
Traceback (most recent call last):
  File "../../bin/transform_mesh", line 39, in <module>
    mesh.points = numpy.apply_along_axis(remap, 1, mesh.points)
AttributeError: 'tuple' object has no attribute 'points'
Makefile:2: recipe for target 'input' failed
make[1]: *** [input] Error 1
```

angus-g:
That might be a meshio version thing? I'm using 3.1.5 locally.

tmbgreaves:
I'll have a poke at that now.

tmbgreaves:
The bionic version is fossilised - reporting as 1.11.7. Can we work with that? At a pinch we could build our own more recent package but as far as possible we're trying to work within the versions available by default from Ubuntu LTS.

tmbgreaves:
Replacing it with the latest version from pip (3.2.10) I get a different error:

Traceback (most recent call last):
  File "../../bin/transform_mesh", line 41, in <module>
    meshio.write(mesh_name + '.msh', mesh)
  File "/usr/local/lib/python3.6/dist-packages/meshio/_helpers.py", line 273, in write
    return interface.write(filename, mesh, *args, **_kwargs)
  File "/usr/local/lib/python3.6/dist-packages/meshio/_gmsh/main.py", line 98, in write
    writer.write(filename, mesh, binary=binary)
  File "/usr/local/lib/python3.6/dist-packages/meshio/_gmsh/msh4_1.py", line 272, in write
    write4_1(filename, mesh, binary=binary)
  File "/usr/local/lib/python3.6/dist-packages/meshio/_gmsh/msh4_1.py", line 320, in write4_1
    _write_nodes(fh, mesh.points, mesh.cells, binary)
  File "/usr/local/lib/python3.6/dist-packages/meshio/_gmsh/msh4_1.py", line 370, in _write_nodes
    assert len(cells) == 1, "Can only deal with one cell type for now"
AssertionError: Can only deal with one cell type for now
Makefile:2: recipe for target 'input' failed

angus-g:
I've rebased this on master, and started to add msh4 reading to gmshtools (to avoid going to meshio due to its fragility).

It's not quite correct: the element numbering makes the `wind_stratification` test fail, and there's no handling of periodic meshes, so `channel_wind_drag_rotated` also fails...

angus-g:
I'm getting the following failures

```
heat-transfer-eqn-channel.xml: PPPPFPPPPPFP
explicit-hyperc-tets-gmsh.xml: PPPPFPPPPPFP
sigma_layer_sphere_parallel.xml: PPPFF
particle_attributes.xml: PPPFPP
explicit-hyperc-tets.xml: PPPPFPPPPPFP
particle_attributes_checkpoint.xml: PPPFPP
swe_dam_break_2d.xml: PFFPPFP
channel_wind_drag_rotated.xml: F
```

I think excluding `channel_wind_drag_rotated`, these are due to differences in mesh generation between gmsh 2 and gmsh 4 (4.5.1 in my case). I was checking out `heat-transfer-eqn-channel`, and it appears to be rotated or reflected in gmsh 4 vs gmsh 2. Ideas?

stephankramer:
Oddly enough, most of these pass for me locally (on your branch, gmsh 4.1.3). Only failures for me are:
* particle_attributes: should check with Chris
* sigma_layer_sphere_parallel: that one has quite a particular mesh, a small segment at the Earth's surface - and it seems to somehow include the centre of the Earth. There might be a better way to do this?
* channel_wind_drag_rotated: gmshtools, which you know about

Not sure what the difference is...

angus-g:
Looks like gmsh 4.1.3 does generate a slightly different mesh to 4.5.1 (but also I think slightly different to 2.16.0), so that might explain the difference in test failures. For me, `spherical_patch` also fails to generate on 4.1.3 with a whole pile of "wrong mesh element size" errors. It's interesting that you didn't hit that?

angus-g:
Thanks @Cmath2, those tests are passing within 1e-29 or something now! A couple of things:
* I guess the test doesn't involve the particle position any more? Do we know why it was so far off in that case?
* I think the same changes will need to be made to `particle_attributes_checkpoint`, which fails in the same way
* It'd be worth actually submitting these changes as a separate PR against master (since this GMSH 4 branch still has the old particle IO). Then I can rebase this back onto master without conflicts.

Cmath2:
@angus-g The test still does involve particle positions in the Salinity field, which is a linear field based on the y co-ordinate.
The problem came from the Temperature field potentially being nonlinear depending on how the mesh is generated. This caused the field to be incorrectly calculated in the .xml file relative to the field carried by particles. (In the simulation the temp field was calculated based on nodal positions then interpolated onto particles, in the .xml file the Temp field was calculated based on particle positions. This lead to the two generating different results).

I'll make a PR for the relevant tests against master now.

Cmath2:
PR is up in #255

tmbgreaves:
I have an immediate use case for these changes  with the Focal builder :-) @stephankramer @Patol75 @jrper might you be able to cast a reviewer's eye over this one again please?

angus-g:
Depending on how immediate you need, you could also pass `-fmt msh2` to gmsh 4 to generate a backward-compatible mesh. I think this should work for almost all of the tests (I fixed a bug in gmsh4 for one problem, so you’ll need a recent enough version)

tmbgreaves:
Thanks @angus-g  - I've just pushed a GMSH v2 build into my container for the moment but as soon as this lands I'll make use of it :-)

tmbgreaves:
@stephankramer might you be able to re-review this PR please, and add any comments?

stephankramer (approved):
Gone through it again. All looks perfect. Many thanks @angus-g Feel free to merge when convenient @tmbgreaves

PR #237, opened 2019-09-17
Fix netcdf_read_free_surface with modern matplotlib

angus-g:
The `bivariate_normal` function was removed from matplotlib's mlab modules [in 3.1.0](https://matplotlib.org/api/prev_api_changes/api_changes_3.1.0.html?highlight=bivariate_normal#matplotlib-mlab-removals). Using `multivariate_normal` from scipy (as of 0.14.0) can replicate the same behaviour: the maximum absolute difference is about 5e-16 (matplotlib 2.1.1, scipy 0.19.1, Python 2.7.15).

Making this change fixes the *netcdf_read_free_surface* test case with recent matplotlib versions.

### DIFF
--- a/tests/netcdf_read_free_surface/height.py
+++ b/tests/netcdf_read_free_surface/height.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python3
 
-from matplotlib.mlab import bivariate_normal
+from scipy.stats import multivariate_normal
 
 def function(position):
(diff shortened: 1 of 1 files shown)


stephankramer:
I'm just giving it a replay: https://jenkins.ese.ic.ac.uk:1080/blue/organizations/jenkins/fluidity/detail/PR-237/5/pipeline

PR #239, opened 2019-11-04
Add option to set vtk FieldData style metadata for TimeValue.

jrper:
For VTK 8.2+ and Paraview 5.5+ this automagically gets used to set the time levels of a file series.

q.v. https://gitlab.kitware.com/vtk/vtk/commit/6e0acf3b773f120c3b8319c4078a4eac9ed31ce1

I'm raising this relatively early, since I suspect there may be some discussion of the right thing to do about this, and whether this should be default on or default off.

### DIFF
--- a/femtools/VTK_interfaces.F90
+++ b/femtools/VTK_interfaces.F90
@@ -103,7 +103,7 @@ module vtk_interfaces
 
 contains
     
-  subroutine vtk_write_state(filename, index, model, state, write_region_ids, write_columns, stat)
+  subroutine vtk_write_state(filename, index, model, state, write_region_ids, write_columns, metadata, stat)
     !!< Write the state variables out to a vtu file. Two different elements
     !!< are supported along with fields corresponding to each of them.
     !!<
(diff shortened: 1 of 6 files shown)


femtools/VTK_interfaces.F90
```diff
@@ -238,6 +241,7 @@ subroutine vtk_write_fields(filename, index, position, model, sfields,&
     type(tensor_field), dimension(:), intent(in), optional :: tfields
     logical, intent(in), optional :: write_region_ids
     logical, intent(in), optional :: write_columns
+    type(scalar_field), dimension(:), optional :: metadata
```
stephankramer, line 244: Hm, this seems to abuse scalar_field as a container for a real array (with a name). A scalar_field is defined on a mesh (function space!), and it suggests we care which mesh. I can see this is a compact way of supplying one or more real arrays of different lengths with names, but if people want to use this in a generic way they really will have to abuse scalar_fields by allocating a %val array that is inconsistent with whatever random mesh they choose to use. Also when first reading this code, I thought it was just writing out an extra field (i.e an actual spatial field), so I think it's better to make it explcit that we're just supplying a real array of arbitrary length.

Since we only use this for time at the moment, I would be happy to just have a `real, intent(in):: time_value` argument (where the name is hardcoded to `TimeValue`). Alternatively you could use an array of `real_vector`s (see femtools/Futils.F90) with a separate array of names.

PR #250, opened 2019-12-09
Fix mpi headers

### LINKED ISSUE #226: MPI linking error while building tools
I'm not really sure what the root of this one is, but during `make fltools`, three tools call for the use of `FLLINKER` (mpif90 in my case). This causes an undefined reference error linking in libfluidity, e.g.
```
../lib/libfluidity.a(C++_Debug.o): In function `MPI::Intracomm::Clone() const':
C++_Debug.cpp:(.text._ZNK3MPI9Intracomm5CloneEv[_ZNK3MPI9Intracomm5CloneEv]+0x2b): undefined reference to `MPI::Comm::Comm()'
```
just to pick an example. For reference, I get this with both OpenMPI 3.1.3 and 4.0.1.

Using the regular `LINKER` (mpicxx) links in the MPI C++ stuff and allows this compile to succeed. I don't have a working configuration without this change so I'm hesitant to suggest anything further.

stephankramer:
This avoid pulling in symbols from C++ MPI interface when including `<mpi.h>` in .cpp files.  We don't use this interface (as it is deprecated) and the spurious symbols may cause issues when linking. Uses `#define MPICH_SKIP_MPICXX` and `OMPI_SKIP_MPICXX` as suggested by petsc's `petscsys.h`

Fixes #226


libvtkfortran/vtkfortran.cpp
```diff
@@ -773,25 +769,7 @@ extern "C" {
...
-        }
-        MPI_Barrier(MPI_COMM_WORLD);
-      }
-    }else{
-#endif
-      _vtkpclose_nointerleave(rank, npartitions);
-#ifdef HAVE_MPI
-    }
-#endif
+    _vtkpclose_nointerleave(rank, npartitions);
```
angus-g: So no interleaving/MPI in libvtkfortran?

include/flmpi.h
```diff
@@ -0,0 +1,45 @@
...
+#ifdef HAVE_MPI
+#if !defined(MPICH_SKIP_MPICXX)
+#  define MPICH_SKIP_MPICXX 1
+#endif
+#if !defined(OMPI_SKIP_MPICXX)
+#  define OMPI_SKIP_MPICXX 1
+#endif
+#if !defined(OMPI_WANT_MPI_INTERFACE_WARNING)
+#  define OMPI_WANT_MPI_INTERFACE_WARNING 0
+#endif
```
angus-g: What's this one for?

drhodrid:
Can this be merged @stephankramer ?

include/flmpi.h
stephankramer: Good point I just copied that randomly from petsc. It seems to supresswarnings about deprecated mpi functions, which I guess we would want to know about.

libvtkfortran/vtkfortran.cpp
stephankramer: It seemed very dodgy and only seems to kick in on npartiions>64000 (!), so I thought it'd be safe to remove.

stephankramer: But as to how this relates to this PR, it means there's no explicit MPI calls in libvtkfortran anymore which means I don't have to introduce this mpi header business for libvtkfortran.  DQMOM is indeed unrelated, so you're right I should have done that in a separate merge.

PR #253, opened 2020-01-20
Prevent unallocated/unassociated actual arguments

angus-g:
In a few places, an array will be allocated or a pointer associated depending on a logical flag. These subroutines then call another subroutine, passing the array/pointer and the flag. However, the dummy argument to this second subroutine doesn't have the `allocatable` or `pointer` attributes. Depending on the flag, we may be passing an unallocated array or unassociated pointer to a bare dummy argument, which is not allowed. To be compliant, we should specify these attributes, even for `intent(in)` dummy arguments, where the allocated/associated status is not allowed to change (this is a Fortran 2003-ism in the case of pointers).

For an example of the pattern:

```fortran
subroutine construct_momentum_element_cg(...)
  real, dimension(:,:,:), allocatable :: dnvfrac_t

  if (multiphase) then
    allocate(dnvfrac_t(...))
  end if

  call add_advection_element_cg(..., dnvfrac_t, ...)
end subroutine construct_momentum_element_cg

subroutine add_advection_element_cg(..., dnvfrac_t, ...)
  real, dimension(:,:,:), intent(in) :: dnvfrac_t
  !                                     ^--- we might pass an unallocated array to this!
  
  if (multiphase) then
    ! do something with dnvfrac_t
  end if
  ...
```

These are just the cases I had to fix to get the short/medium tests to run with the Intel compiler with checking enabled, so I'm not sure if it's exhaustive. Ideally this'd be statically analysed somehow

### DIFF
--- a/assemble/Momentum_CG.F90
+++ b/assemble/Momentum_CG.F90
@@ -1614,7 +1614,7 @@
       type(scalar_field), intent(in) :: nvfrac
       real, dimension(ele_loc(u, ele), ele_ngi(u, ele), u%dim), intent(in) :: du_t
       real, dimension(ele_loc(u, ele), ele_ngi(u, ele), u%dim), intent(in) :: dug_t
-      real, dimension(:, :, :), intent(in) :: dnvfrac_t
+      real, dimension(:, :, :), allocatable, intent(in) :: dnvfrac_t
       real, dimension(ele_ngi(u, ele)), intent(in) :: detwei
       real, dimension(u%dim, u%dim, ele_ngi(u,ele)) :: J_mat, diff_q
       real, dimension(u%dim, u%dim, ele_loc(u, ele), ele_loc(u, ele)), intent(inout) :: big_m_tensor_addto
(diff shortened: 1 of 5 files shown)


angus-g:
Maybe an alternative is to make these `optional`?

> A dummy argument [...] is not present if the dummy argument
> * [...]
> * does not have the `ALLOCATABLE` or `POINTER` attribute, and corresponds to an actual argument that
>   - has the `ALLOCATABLE` attribute and is not allocated, or
>   - has the `POINTER` attribute and is disassociated

stephankramer:
Interesting, I didn't know that. That might actually simplify some code in other places - petsc_solve_state comes to mind, which has a whole bunch of `if (associated)` cases. I presume it means the pointer has to be at least initialised though, I'm guessing? I interpret 'disassociated' as `.not. associated(...)` which is not defined when the pointer hasn't been initialised (e.g. nullified).

In any case, I'm happy with your fixes - but that would be a good suggestion for how to do this in the future.

angus-g:
That's right, a non-initialised pointer has the _undefined_ association status, separate from _associated_ and _disassociated_. At least in the case of `petsc_solve_state`, it seems like the pointers are either explicitly associated or nullified, so it could work there.

stephankramer:
@angus-g: should this go in now?

angus-g:
I think these are all the cases I caught while running the test suite, so this should be good to go.

Issue #236, opened 2019-08-05
Do we want Statplot to depend onto classes defined elsewhere?


Patol75:
I am just wondering if it makes sense to have Statplot using the Stat class defined in `./python/fluidity/diagnostics/fluiditytools.py`, which depends on the stat_parser class defined in `./python/fluidity_tools.py`, and the JoinStat class defined in `./python/fluidity/diagnostics/fluiditytools.py` as well? The reason for this is that all that is needed in Statplot is to be able to read through potentially multiple statfiles and display their information on a single graph. I have an attempt (attached below, function ReadData) which is working for multiple statfiles that have exactly the same fields defined (restarted checkpoint). Implementing this solution would make Statplot a standalone. Please let me know your opinion on this.

EDIT: I have created a branch with the newest version https://github.com/FluidityProject/fluidity/commit/2abc6758e0e71f33ad0adde86ecf1476ae1db028

stephankramer:
I suggest you open a PR instead? Does it work for binary stat/detectors files? If it's very limited code duplication then it would be nice to have it stand-alone, yes - but I'm less keen if this would get any bigger.

PR #254, opened 2020-01-24
Statplot as a standalone

Patol75:
I had already proposed this idea in this now-closed [issue](https://github.com/FluidityProject/fluidity/issues/236) where I tried to show that many files in the Fluidity tree are currently involved in the file-reading process of statplot.

I believe that the current situation is too complex for what it should be and therefore I hereby propose a standalone version of statplot.

Changes compared to the master branch are summarized in the two commit messages. Instructions relative to the implemented keyboard shortcuts are available through `statplot.py --help`.

I am happy to modify further and implement new features if requested.

### DIFF
--- a/tools/statplot.py
+++ b/tools/statplot.py
@@ -53,16 +53,19 @@ class StatplotWindow(Gtk.Window):
         hbox.set_homogeneous(True)
         self.vbox.pack_end(hbox, False, False, 0)
         self.entries, self.values = self.ReadData()
-        self.xCombo = Gtk.ComboBoxText.new_with_entry()
-        self.xCombo.set_wrap_width(3)
-        self.yCombo = Gtk.ComboBoxText.new_with_entry()
-        self.yCombo.set_wrap_width(3)
-        self.PopulateCombo('load')
+        self.store = Gtk.ListStore.new([str])
+        for entry in self.entries:
+            self.store.append([entry])
+        self.xCombo = Gtk.ComboBox.new_with_model_and_entry(self.store)
+        self.yCombo = Gtk.ComboBox.new_with_model_and_entry(self.store)
+        self.ComboParams(self.xCombo)
+        self.ComboParams(self.yCombo)
+        self.InitCombo('load')
         self.InitCompletion(self.xCombo)
         self.InitCompletion(self.yCombo)
-        self.xCon = self.xCombo.connect('changed', self.ComboChangedX)
+        self.xCon = self.xCombo.connect('changed', self.ComboChanged)
         self.xCombo.connect('key-release-event', self.ReleaseFocus)
-        self.yCon = self.yCombo.connect('changed', self.ComboChangedY)
+        self.yCon = self.yCombo.connect('changed', self.ComboChanged)
         self.yCombo.connect('key-release-event', self.ReleaseFocus)
         hbox.pack_start(self.xCombo, True, True, 0)
         hbox.pack_start(self.yCombo, True, True, 0)
@@ -76,17 +79,19 @@ class StatplotWindow(Gtk.Window):
         self.toolbar = NavigationToolbar(self.canvas, self)
         self.vbox.pack_start(self.toolbar, False, False, 0)
 
-    def ComboChangedX(self, widget):
-        if self.xCombo.get_active_text() in self.entries:
+    def ComboChanged(self, combo_box):
+        activeText = combo_box.get_child().get_text()
+        if activeText in self.entries:
+            combo_box.set_active(self.entries.index(activeText))
             self.PlotData('update', self.PlotType, 'linear',
                           self.ax.get_yscale())
             self.canvas.grab_focus()
 
-    def ComboChangedY(self, widget):
-        if self.yCombo.get_active_text() in self.entries:
-            self.PlotData('update', self.PlotType, self.ax.get_xscale(),
-                          'linear')
-            self.canvas.grab_focus()
+    def ComboParams(self, comboBox):
+        comboBox.set_entry_text_column(0)
+        comboBox.set_margin_left(100)
+        comboBox.set_margin_right(100)
+        comboBox.set_wrap_width(3)
 
     def FormatAxis(self, ax2fmt, scale):
         if ax2fmt == 'x':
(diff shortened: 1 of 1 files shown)


tools/statplot.py
```diff
@@ -148,46 +155,60 @@ def FormatAxis(self, ax2fmt, scale):
...
-                self.Data = self.xData
+                curData = self.xData
             elif key == 'y':
                 self.get_scale = self.ax.get_yscale
-                self.Data = self.yData
-            if self.get_scale() == 'linear' \
-                    and self.Data[self.Data != 0].size == 0:
+                curData = self.yData
+            if (self.get_scale() == 'linear'
+                    and curData[curData != 0].size == 0):
```
angus-g: I'd suggest changing this to `(curData == 0).any()` just to make the intention clearer (took me a while to figure out what this was actually doing!)

tools/statplot.py
```diff
@@ -204,14 +225,28 @@ def KeyPressed(self, widget, event):
...
-        self.xData = self.entries[self.xField]
-        self.yData = self.entries[self.yField]
-        if numpy.unique(self.xData).size == 1 \
-                and numpy.unique(self.yData).size == 1:
+        if len(self.values.shape) == 1:
+            self.xData = self.values[self.xCombo.get_active()]
+            self.yData = self.values[self.yCombo.get_active()]
+        else:
+            self.xData = self.values[:, self.xCombo.get_active()]
+            self.yData = self.values[:, self.yCombo.get_active()]
```
angus-g: I think you can tidy this up with:

```python
self.xData = self.values[..., self.xCombo.get_active()]
```

Should work for any number of dimensions

tools/statplot.py
```diff
@@ -237,50 +272,74 @@ def PlotData(self, action, type, xscale, yscale):
...
-            else:
-                self.xCombo.set_active(0)
-                self.yCombo.set_active(1)
-        elif action == 'reload':
-            self.xCombo.set_active_iter(prevIterX)
-            self.yCombo.set_active_iter(prevIterY)
+    def ReadData(self):
+        def GatherEntries(stat2read):
+            print('Reading ' + stat2read)
+            entries = numpy.array([])
```
angus-g: I think this might be clearer as a regular Python list, then later on we'd only need

```python
entries.append(...)
```

To support the argsort indexing later, it can be converted to a numpy array from the list

tools/statplot.py
```diff
@@ -237,50 +272,74 @@ def PlotData(self, action, type, xscale, yscale):
...
+                        info = etree.fromstring(line)
+                        if info.get('statistic') == 'value':
+                            entries = numpy.append(entries, info.get('name'))
+                        elif all([x in info.keys() for x in
+                                  ['components', 'material_phase']]):
+                            for i in range(int(info.get('components'))):
+                                entries = numpy.append(
+                                    entries, info.get('material_phase') + '%'
+                                    + info.get('name') + '%'
+                                    + info.get('statistic') + f'%{i}')
```
angus-g: I think this (and the following few times this is done) would be clearer as a full format strings (using the other suggestion above). Maybe something like:

```python
for i in range(...):
  entries.append("{e[material_phase]}%{e[name]}%{e[statistic]}%i".format(e=info.attrib, i=i))
```

tools/statplot.py
```diff
@@ -237,50 +272,74 @@ def PlotData(self, action, type, xscale, yscale):
...
-            if failcount == 4:
-                stats.append(fluidity_tools.Stat(filename))
-        if len(stats) == 1:
-            return stats[0]
-        else:
-            return fluidity_tools.JoinStat(*stats)
+        if self.statfile[0].endswith('stat'):
+            entries, num = GatherEntries(self.statfile[0])
+            values = numpy.genfromtxt(self.statfile[0], skip_header=num + 1)
+            values = values.T[numpy.argsort(entries)].T
```
angus-g: Can you just index along the second dimension instead of double-transposing (`values[:, numpy.argsort(entries)]`)?

tools/statplot.py
```diff
@@ -237,50 +272,74 @@ def PlotData(self, action, type, xscale, yscale):
...
+                        valuesTemp = valuesTemp.T[numpy.argsort(entriesTemp)].T
+                        values = numpy.vstack((values, valuesTemp))
+                    else:
+                        sys.exit('Statfiles entries do not match')
+        elif self.statfile[0].endswith('detectors'):
+            entries = GatherEntries(self.statfile[0])[0]
+            values = numpy.fromfile(self.statfile[0] + '.dat',
+                                    dtype=numpy.float64)
+            ncols = entries.size
+            nrows = numpy.floor(values.size / ncols).astype(int)
```
angus-g:
```suggestion
            nrows = values.size // ncols
```

tools/statplot.py
Patol75: Thanks for that, and sorry for the headache! I have used all instead of any because matplotlib can automatically handle cases with only a few zeros.

tools/statplot.py
Patol75: Thanks for showing an example use of Ellipsis objects, never really tried to understand what they are before!

tools/statplot.py
Patol75: Agree the Numpy append syntax is quite verbose; fixed.

tools/statplot.py
Patol75: Cheers, looks cleaner that way!

tools/statplot.py
Patol75: True, thank you.

tools/statplot.py
Patol75: Just needed an additional int() as values.size can be a float, but good idea indeed.

angus-g: The size can be a float!?

Patol75: Woops, memory confusion, it is the quotient that can be a float, so indeed no need for int(). Thanks for paying attention.

Patol75:
Thanks @angus-g. Can I hit the merge button @stephankramer or you would still like to have a look at it?

drhodrid:
Merge away!

On 14 Feb 2020, at 3:19 pm, Thomas Duvernay <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks @angus-g<https://github.com/angus-g>. Can I hit the merge button @stephankramer<https://github.com/stephankramer> or you would still like to have a look at it?

PR #257, opened 2020-02-04
Add array, vector and tensor capability to particle attributes

angus-g:
Currently, particle attributes can only be a scalar value. This PR adds the ability to specify vector and tensor attributes. In addition, any attribute can also be an array (e.g. an array of tensors).

Along with this addition is a fairly significant refactor of the particle code. In particular, the reading of attributes from the configuration is centralised and stored along with the particles, instead of being repeated. The Python interface code has also been refactored to remove code duplication and hopefully make it a bit easier to understand the memory management when transitioning between the languages.


femtools/Detector_Tools.F90
```diff
@@ -643,65 +742,324 @@ subroutine set_particle_attribute_from_python(attributes, positions, nparts, fun
...
+
+    deallocate(field_vals)
+    deallocate(old_field_vals)
+  end subroutine set_particle_scalar_attribute_from_python_fields
+
+  !> Given the particle position and time, evaluate the specified python function for a group of particles
+  !! specified in the string func at that location.
+  subroutine set_particle_vector_attribute_from_python(attributes, positions, nparts, natt, func, time, dt, is_array)
+    !! Attribute values to set
+    real, dimension(:,:), intent(out) :: attributes
```
stephankramer: I was expecting a rank-3 array here: dim X natt X nparts

femtools/Particles.F90
```diff
@@ -57,33 +56,102 @@ module particles
...
+    !> Model state structure
     type(state_type), dimension(:), intent(in) :: state
-    logical, intent(in), optional :: global !! use global/parallel picker queries to determine particle elements
-    logical, intent(in), optional :: from_flredecomp !! change behaviour if we're being called by flredecomp
-    integer, intent(in), optional :: number_of_partitions !! number of processes to use for reading
+    !> Use global/parallel picker queries to determine particle elements?
+    logical, intent(in), optional :: global
+    !> Whether we're being called by flredecomp, this changes behaviour around
+    !! parallel access to particle files
+    logical, intent(in), optional :: from_flredecomp
```
stephankramer: In general I prefer if logicals just say what they do, rather than "changes behaviour if called within a certain context" - if necessary with multiple logicals representing the different bits that are switched on or off. Fluidity used to be littered with this sort of context-specific logic that tends to propagate through different layers as well. Instead, every bit of code should make sense on its own.

schemas/fluidity_options.rnc
```diff
@@ -3718,6 +3787,107 @@ generic_particle_attribute_choice =
...
+      element constant {
+         real_dim_vector
+      }|
+      ## Python function prescribing real input. Functions should be of the form:
+      ##
+      ##  def val(X, t, dt):
+      ##     # Function code
+      ##     return # Return value
+      ##
+      ## where X is a tuple of length geometry dimension.
```
stephankramer: X and the return value is a tuple of length...

schemas/fluidity_options.rnc
```diff
@@ -3718,6 +3787,107 @@ generic_particle_attribute_choice =
...
+      element constant {
+         real_dim_tensor
+      }|
+      ## Python function prescribing real input. Functions should be of the form:
+      ##
+      ##  def val(X, t, dt):
+      ##     # Function code
+      ##     return # Return value
+      ##
+      ## where X is a tuple of length geometry dimension.
```
stephankramer: And the return value is dim X dim tensor in the form of a list of lists (?) or anything compatible with a dim X dim numpy array I presume?

angus-g:
Just noting here (from an email exchange) that there are a few new test failures introduced by the changes in the Python embedding, particularly when setting a tensor field.

angus-g:
I was doing `result[ii*(dim[0] + dim[1]) ...` rather than `dim[0]*dim[1]` which of course I wasn't hitting on 2D cases and was very hard to spot as obviously wrong! Still a few more tweaks to push through on this branch.

femtools/Detector_Data_Types.F90
```diff
@@ -37,22 +37,43 @@ module detector_data_types
...
+  integer, parameter :: STATIC_DETECTOR=1, LAGRANGIAN_DETECTOR=2
 
   type stringlist
-     !!< Container type for a list of strings.
-     character(len=FIELD_NAME_LEN), dimension(:), pointer :: ptr
+    !!< Container type for a list of strings.
+    character(len=FIELD_NAME_LEN), dimension(:), pointer :: ptr
   end type stringlist
 
+  type attr_names_type
```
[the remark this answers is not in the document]
angus-g: I do like your suggestion, but I think I went with the more split-out types to avoid some redundancy in the objects we pass around. I guess the way I've written it, you pass a few objects like `names` and `vals` that encapsulate scalar, vector and tensor components. Your way means that we pass around the scalar, vector and tensor versions of a single descriptor.

It would be a big change to rewrite with a new data structure, so it's my mistake for picking something a bit awkward!

femtools/Detector_Tools.F90
```diff
@@ -612,24 +633,102 @@ subroutine set_detector_coords_from_python(values, ndete, func, time)
...
+      end do
+
+      field_idx = field_idx + dim**2
+    end do
+  end subroutine evaluate_particle_fields
+
+  !> Given the particle position and time, evaluate the specified python function for a group of particles
+  !! specified in the string func at that location.
+  subroutine set_particle_scalar_attribute_from_python(attributes, positions, nparts, natt, func, time, dt, is_array)
+    real, dimension(:,:), intent(out) :: attributes
```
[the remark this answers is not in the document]
angus-g: > Is this `natt X nparts`
Yes -- I've clarified that

> Since this is the fortran interface, do we need to provide nparts and natt seperately? Can we note use size(attributes,1) and size(attributes,2)?

Sure, that's a good idea.

femtools/Detector_Tools.F90
angus-g: All the attributes on a particle list are stored in a 2D array `(total att x nparts)`, so while I think it could make more sense to pass a 3D array into this function, it's a little more awkward.

femtools/Detector_Tools.F90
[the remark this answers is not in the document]
angus-g: Actually looking through the following vector and tensor functions, where `natt` is different from `size(attributes,1)`, it would be more consistent to still specify this. We can still drop `nparts` though.

schemas/fluidity_options.rnc
angus-g: Whoops, good catch! We ask numpy for `PyArray_DIMS` after `PyArray_ContiguousFromObject`, so I think your presumption is correct.

angus-g:
I think I've handled the "few more tweaks", made the tests passed, and addressed your previous feedback @stephankramer. We would probably do with re-enabling the `particle_attribute_array` test after making it do something sensible (ideas @Cmath2?), and we might be good to finally tick this one off!?

Cmath2:
particle_attribute_array test has been setup and is passing locally on my system.

angus-g:
Awesome, thanks @Cmath2! Passing locally here too. Hopefully we get through Jenkins!

drhodrid:
@stephankramer - this pull request is now green on Jenkins and Angus has been through to address your previous comments. Any chance you could take one last look?

drhodrid:
@tmbgreaves - any chance you could give jenkins a poke here? Seems to be misbehaving... (unless I'm missing something)...

stephankramer:
I've given it a replay. Looks like a temporary outage of dockerhub

stephankramer:
@angus-g or @Cmath2 : this seems to now conflict with master. Could you merge in master and resolve the conflict, please? I think all issues have been addressed, so I'm ready to push that green button!

angus-g:
Okay, I've rebased onto the latest master. I re-ran `update_options.py` on the test `flml` files that had been touched recently, and the only other change was the removal of `timestep` from the `move_lagrangian_detectors()` call, as that lost that parameter in the particle speedup PR.

PR #258, opened 2020-02-13
Speedup particle advection

stephankramer:
Nothing controversial, I think. Speed up of particle advection code by:
* speedup local_coords() by reusing inverse jacobian that is already being cached in transform_elements
* only call local_coords() half the number of times by trusting the local coordinates that are stored on the detector
* speed up eval_field by explcitly writing out the P2 case rather than relying on the generic code using polynomials
* only call eval_field() half the number of times by theta-weighting the velocity beforehand

Finally, unrelated to particle advection: in a lot of places instead of calling local_coords_node, which evaluates the basis functions for one node, use the now optimized local_coords_interpolation which evaluates all nodes of an element in one go. This should speed up some interpolation code and places where we do remap_field.


Patol75:
From the Intel Profiler, I think the run time dropped from some 800 seconds to some 300, which is pretty awesome. This, unfortunately, did not translate into an overall noticeable speed-up in a 3-D simulation like the ones I am running.

Cmath2:
> From the Intel Profiler, I think the run time dropped from some 800 seconds to some 300, which is pretty awesome. This, unfortunately, did not translate into an overall noticeable speed-up in a 3-D simulation like the ones I am running.

The speedup only affects particle advection routines. For large 3-D simulations like yours, particle advection only accounts for a fraction of the total simulation time.

Overall particle advection is on the order of 65-75% faster with these changes

angus-g:
I guess this is ready to merge, but maybe we'd like to throw Jenkins at it first just to be sure (@tmbgreaves?)

tmbgreaves:
Jenkins appears to have had a power cut over the weekend and the workers had come up without docker. I've restarted the docker daemons and kicked the runs to restart - the error it just gave was a spurious one (broken docker). Hopefully a real status will emerge in a few hours.

angus-g:
Looks like a real failure now, in the unit tests

angus-g:
Also looks like a potential tolerance failure on `detectors_mesh_movement`:

```
detectors_mesh_movement: Assigning D1_tracer = [ 0.1  0.1  0.1  0.1  0.1  0.1  0.1  0.1  0.1  0.1]
```

The test is:

```python
assert abs(D1_tracer - 0.1).all() < 1e-12
```

stephankramer:
> The test is:
> 
> ```python
> assert abs(D1_tracer - 0.1).all() < 1e-12
> ```

That test is of course completely wrong! But it's funny how you can stare blind at it: the all method is applied to the float array abs(D1_tracer - 0.1); So first it converts the floats to logical: only where the difference is exactly zero, you get False otherwise True. Then it take all(), which gives False, if any of the differences are exactly zero. That result is compared with a float, so the bool is converted back to float again, giving 0.0<1e-12 if any of the differences are exactly zero and the test passes. What was meant was of course `(abs(D1_tracer - 0.1) < 1e-12).all()`
Hopefully fixed now :crossed_fingers:

PR #259, opened 2020-02-14
Surface l2 norm

stephankramer:
Add surface_l2norm option
    
Same as surface_integral, except computes sqrt(\int_surface field^2 dx)
Also clean some awkward code in surface_integrals module.

Use this new diagnostic in all the various viscous free surface tests.


drhodrid:
@cianwilson - any chance you could take a quick peek at this?

cianwilson:
@stephankramer : think you may have pushed the validation fixed to FluidityStokes but this pull request originates in FluidityProject.

stephankramer:
Thanks, you're right - have pushed it here now and removed the surface-l2-norm branch in FluidityStokes to avoid confusion.

Issue #266, opened 2020-04-17
PETSc has conflict variable is_sorted with Fluidity


gnikit:
So PETSc(v3.13) introduced 6 months ago an enum that contains a variable `IS_SORTED = 0` in `petscis.h`. That creates a compiler conflict with Fluidity's procedure interface `is_sorted`. Is it possible to rename the interface to `is_it_sorted` or something else to resolve the conflict?

I know that officially PETSc 3.9+ is not supported but this is the only thing preventing Fluidity master from compiling with the latest version of PETSc.

stephankramer:
Go for it. You could also just remove the generic interface and directly call `sparsity_is_sorted` in the unittest (doesn't look like it's used anywhere else).

PR #267, opened 2020-04-17
Remove is_sorted allowing configure with PETSc3.13

gnikit:
PETSc 3.13 added is_sorted as an enum hence causing
a compiler conflict.

Fixes issue #266.


GNikit:
I feel that I should mention that this will compile up to PETSc version 3.13 and more specifically commit `d91aa2297d6f31c2d2f0e7ba09a0622130109be4` since after that commit bug fixes where introduced to `PCBJacobiGetSubKSP` causing some compilation issues on the fluidity side of things.

PR #275, opened 2020-08-06
Bug fixes in statplot

Patol75:
Hi everyone,
I am addressing a few bugs in statplot; the main points are listed in the commit message. There should not be anything controversial, hopefully.
Please let me know if anything is off.
Thomas

### DIFF
--- a/tools/statplot.py
+++ b/tools/statplot.py
@@ -56,11 +56,9 @@ class StatplotWindow(Gtk.Window):
         store = Gtk.ListStore.new([str])
         for entry in self.entries:
             store.append([entry])
-        self.xCombo = Gtk.ComboBox.new_with_model_and_entry(store)
-        self.yCombo = Gtk.ComboBox.new_with_model_and_entry(store)
-        self.ComboParams(self.xCombo)
-        self.ComboParams(self.yCombo)
-        self.InitCombo('load')
+        self.xCombo = self.CreateCombo(store)
+        self.yCombo = self.CreateCombo(store)
+        self.InitCombo()
         self.InitCompletion(self.xCombo)
         self.InitCompletion(self.yCombo)
         self.xCombo.connect('changed', self.ComboChanged)
@@ -92,11 +90,13 @@ class StatplotWindow(Gtk.Window):
             self.canvas.grab_focus()
 
     @staticmethod
-    def ComboParams(comboBox):
+    def CreateCombo(store):
+        comboBox = Gtk.ComboBox.new_with_model_and_entry(store)
         comboBox.set_entry_text_column(0)
         comboBox.set_margin_left(100)
         comboBox.set_margin_right(100)
         comboBox.set_wrap_width(3)
+        return comboBox
 
     def FormatAxis(self, ax2fmt, scale):
         if ax2fmt == 'x':
@@ -160,19 +160,15 @@ class StatplotWindow(Gtk.Window):
                                             color='xkcd:black')
         self.fig.set_tight_layout(True)
 
-    def InitCombo(self, action, prevIterX=None, prevIterY=None):
-        if action == 'load':
-            if 'ElapsedTime' in self.entries:
-                iterX = self.xCombo.get_model().get_iter(
-                    self.entries.index('ElapsedTime'))
-                self.xCombo.set_active_iter(iterX)
-                self.yCombo.set_active(0)
-            else:
-                self.xCombo.set_active(0)
-                self.yCombo.set_active(1)
-        elif action == 'reload':
-            self.xCombo.set_active_iter(prevIterX)
-            self.yCombo.set_active_iter(prevIterY)
+    def InitCombo(self):
+        if 'ElapsedTime' in self.entries:
+            iterX = self.xCombo.get_model().get_iter(
+                self.entries.index('ElapsedTime'))
+            self.xCombo.set_active_iter(iterX)
+            self.yCombo.set_active(0)
+        else:
+            self.xCombo.set_active(0)
+            self.yCombo.set_active(1)
 
     @staticmethod
     def InitCompletion(comboBox):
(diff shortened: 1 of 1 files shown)


tools/statplot.py
```diff
@@ -183,51 +189,62 @@ def KeyPressed(self, widget, event):
...
-                axMax = max(abs(curData))
-                if axMin != axMax and min(curData) < 0:
-                    scale = 'symlog'
-                elif axMin != axMax:
-                    scale = 'log'
-                else:
+            if self.values.ndim == 1:
+                warnings.warn('Insufficient data available to turn on '
+                              + 'logarithmic scale', stacklevel=2)
+            else:
```
angus-g, line 201: You could get rid of a level of indentation here with an early return (de-indent the `else` block). Also, fun fact: you don't actually need to explicitly add the string parts together. I think it looks cleaner without, but I'll defer to your judgement!

```suggestion
            if self.values.ndim == 1:
                warnings.warn('Insufficient data available to turn on '
                              'logarithmic scale', stacklevel=2)
                return
```

tools/statplot.py
```diff
@@ -183,51 +189,62 @@ def KeyPressed(self, widget, event):
...
+                elif key == 'y':
+                    self.get_scale = self.ax.get_yscale
+                    curData = self.yData
+                if (self.get_scale() == 'linear'
+                        and (curData == 0).all()):
                     warnings.warn('Change to logarithmic scale denied: the '
-                                  + 'selected variable for the ' + key + ' '
-                                  + 'axis is a constant.', stacklevel=2)
+                                  + 'selected variable for the ' + key
+                                  + ' axis is null.', stacklevel=2)
```
angus-g, line 212: I don't know what version of Python we're targeting, but maybe this would be nicer with a format-string?

```suggestion
                                  f'selected variable for the {key}'
                                  ' axis is null.', stacklevel=2)
```

tools/statplot.py
```diff
@@ -183,51 +189,62 @@ def KeyPressed(self, widget, event):
...
-                scale = 'linear'
-            self.FormatAxis(key, scale)
-            self.fig.canvas.draw()
+                elif (self.get_scale() == 'linear'
+                      and max(abs(curData)) / min(abs(curData)) < 10):
+                    warnings.warn('Change to logarithmic scale denied: the '
+                                  + 'selected variable for the ' + key
+                                  + ' axis has a range of variation smaller '
+                                  + 'than one order of magnitude.',
+                                  stacklevel=2)
```
angus-g, line 220: More string formatting?

tools/statplot.py
```diff
@@ -183,51 +189,62 @@ def KeyPressed(self, widget, event):
...
-            self.PlotType = 'line' if self.PlotType == 'marker' else 'marker'
-            self.PlotData('update', self.PlotType, self.ax.get_xscale(),
-                          self.ax.get_yscale())
+            if self.values.ndim == 1:
+                warnings.warn('Insufficient data available to turn on line '
+                              + 'display', stacklevel=2)
+            else:
+                self.PlotType = 'marker' if self.PlotType == 'line' else 'line'
+                self.PlotData('update', self.PlotType, self.ax.get_xscale(),
+                              self.ax.get_yscale())
```
angus-g, line 247: Early return pattern here too?

```suggestion
            if self.values.ndim == 1:
                warnings.warn('Insufficient data available to turn on line '
                              + 'display', stacklevel=2)
                return

            self.PlotType = 'marker' if self.PlotType == 'line' else 'line'
            self.PlotData('update', self.PlotType, self.ax.get_xscale(),
                          self.ax.get_yscale())
```

tools/statplot.py
```diff
@@ -277,11 +293,11 @@ def GatherEntries(stat2read):
                                   ['components', 'material_phase']]):
                             for i in range(int(info.attrib['components'])):
                                 entries.append('{e[material_phase]}%{e[name]}'
-                                               '%{e[statistic]}%i'
+                                               '%{e[statistic]}%{i}'
```
angus-g, line 296: Good catch!

tools/statplot.py
```diff
@@ -294,13 +310,13 @@ def GatherEntries(stat2read):
...
             return numpy.asarray(entries), num
 
-        if self.statfile[0].endswith('stat'):
-            entries, num = GatherEntries(self.statfile[0])
-            values = numpy.genfromtxt(self.statfile[0], skip_header=num + 1)
-            values = values[:, numpy.argsort(entries)]
+        if statfile[0].endswith('stat'):
+            entries, num = GatherEntries(statfile[0])
+            values = numpy.genfromtxt(statfile[0], skip_header=num + 1)
+            values = values[..., numpy.argsort(entries)]
```
angus-g, line 316: Nice, I prefer ellipsis indexing

tools/statplot.py
```diff
@@ -236,34 +253,33 @@ def KeyPressed(self, widget, event):
     def PlotData(self, action, type, xscale, yscale):
         self.xData = self.values[..., self.xCombo.get_active()]
         self.yData = self.values[..., self.yCombo.get_active()]
-        if (numpy.unique(self.xData).size == 1
-                and numpy.unique(self.yData).size == 1):
-            type = 'marker'
         if action == 'create':
-            self.statplot, = self.ax.plot(self.xData, self.yData, linewidth=2,
-                                          color='xkcd:light purple')
+            self.statplot, = self.ax.plot(self.xData, self.yData)
```
angus-g, line 257: This `PlotData` method would be a little easier to read with a couple of blank lines (before the `if` lines and after the `elif` lines, for example)

Patol75:
Thank you for the comments @angus-g. I have incorporated them all and pushed alongside with two minor updates, as detailed in the commit message. Thank you for your time on this. :)

PR #276, opened 2020-08-11
Remove logic preventing momentum CG weak bc assembly unless advection…

stephankramer:
… or continuity were integrated by parts.

To get into the momentum weak bc assembly in Momentum_CG you needed to either integrate advection or continuity by parts.  This
meant that flux bcs weren't assembled unless these criteria were met, which is clearly a bug.  For now I'm just removing that
logic but this should be improved as now all surfaces will have transform_to_physical called regardless of whether they are having
weak bcs applied.

### DIFF
--- a/assemble/Momentum_CG.F90
+++ b/assemble/Momentum_CG.F90
@@ -759,65 +759,62 @@
 
       ! ----- Surface integrals over boundaries -----------
       
-      if((integrate_advection_by_parts.and.(.not.exclude_advection)).or.&
-           (integrate_continuity_by_parts)) then
-         allocate(velocity_bc_type(u%dim, surface_element_count(u)))
-         call get_entire_boundary_condition(u, &
-           & (/ &
-             "weakdirichlet ", &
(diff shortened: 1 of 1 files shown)


stephankramer:
I'm happy for this to go in. I don't think I care to much about the saving by avoiding transform to physical in some cases, as it's only for surface elements, so I doubt it's going to be significant. I prefer to keep things simple.

cianwilson:
OK, I'm not eligible to review this obviously but glad you think it's OK.  I'm happy to try to implement something more "efficient" if needed.  Don't think it would be that hard.  Thanks @stephankramer !

stephankramer:
Well according to github you are (eligible to review), and I am not (as I submitted the PR). So I'll just ask @drhorid to ~~do a thorough review~~ hit that approve button then :-)

cianwilson:
The weird and wonderful logic of github.  Thanks @drhodrid and @stephankramer for getting my commit in with zero responsibility on me ;-)

PR #278, opened 2020-08-11
Big L2 nullspace projection

stephankramer:
This performs a big L2 projection after the final velocity correction if a nullspace has been specified for Velocity, i.e. it ensure that the final velocity is L2 orthogonal to the specified nullspace. This is to avoid spurious rotations (or translations) that may occur with variable resolution. During the (petsc) solves the nullspace is projected out in the small l2-norm. However if for example the mesh resolution is biased where there are e.g. the same number of nodes with clockwise rotation as there are nodes with counter-clockwise rotation, but the nodes with clockwise rotation are very much concentrated in small area, then in an L2-integrated sense the solution still has a net counter-clockwise rotation (as the same number of associated nodes represent more area).

A possibly controversial aspect: I did not introduce any option. It now always does the L2 projection if Velocity has a nullspace. My reason (other than reluctance to add yet more options) is that I think this is the correct default - so if people really wanted to use the old default (i.e. stick with the small l2 projected solution) we could add an option for that. In practice, nullspaces are only used in Stokes applications anyway  and again I think there it is just the right thing to do.

Second point is that this isn't implemented for scalar solves. The (lazy) reason there is that it mucks quite a bit with solver interfaces, as we need a positions field for L2 projection, and generally we don't really care so much. For pressure we only care about its gradient, and for a scalar diffusion problem it could just be done as a post-processing step. This could be added in the future as well.

Final confession: I seem to have repurposed `dot_product(fieldA, fieldB)` to compute the L2 product rather than the l2-product which it did before. I presume this means it wasn't actually used previously - it certainly wasn't parallel safe previously....

### DIFF
--- a/assemble/Momentum_Equation.F90
+++ b/assemble/Momentum_Equation.F90
@@ -1828,8 +1828,9 @@
          type(block_csr_matrix), dimension(:), intent(inout) :: subcycle_m
          type(vector_field), dimension(:), intent(inout) :: subcycle_rhs
 
-         ! Slope limiter variables:
          integer :: d
+         type(vector_field), pointer :: x
+         type(vector_field) :: u_positions
          type(scalar_field) :: u_cpt
 
          ewrite(1,*) 'Entering finalise_state'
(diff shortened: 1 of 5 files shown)


stephankramer:
> Obviously the build failure needs fixing first but I'm curious if there are any tests that depend on this? Are they in another branch that's waiting for a few other merges?

The final nullspace removal will occur in all current tests that have a velocity nullspace - but yes, I'm currently assuming (hoping? fingers crossed) that these do not depend on it. The convergence tests in analytical-benchmarks-remeaned will depend on the correct nullspace removal, so this will be properly tested (in master).

> Just to check, is it the distinction between (l2) removing in a Krylov solver and (L2) removing as post-processing that means the former needs to be ortho-normalized but the latter doesn't seem to?

The only reason petsc requires you to ortho-normalise is so that they don't have to divide by n_dot_n every projection (in every iteration) - so it's an efficiency thing - whereas we only do the projection once. I suppose you could cache L2 normalized null space solutions between timesteps, but I don't think it's worth it.

PR #279, opened 2020-08-12
Layered extrusion

stephankramer:
This add the ability to specify mesh extrusion in stages (layers). Each layers can specify its own regions, sizing_function, etc. In particular this allows the specification of internal boundaries between the layers by setting top_ and bottom_surface_id in the top layer, where top_surface_id specifies the surface id of the external top boundary, and bottom_surface_id the id of the internal boundary between the top and the next layer, in each subsequent layer bottom_surface_id can be specified to specify the internal boundary between this and the next layer. In the bottom layer the bottom_surface_id refers to the external bottom boundary. bottom_depth in each layer is measured as a positive depth from the top at z=0.

The specification of layers is now optional and should have no consequence for exisiting (single layer) extruded cases.

A single simple test is included. The analytical-benchmarks will add a number of additional cases.

Note that the history is a little convoluted, as this started out with layers inside regions, which then changed to regions inside layers with both changing all existing extruded tests. The existing single layer specification with regions directly under extrude/ is now also supported though, so these test cases have been reverted. This also means this PR no longer includes an updateflml script as no update to existing flmls is required.


stephankramer:
Thanks @cianwilson ! I addressed your comments. I think you're right and they are copied from the column ids diagnostic - that one does seem to have some sort of check already (which looks believably right)

stephankramer:
And rebased

PR #282, opened 2020-08-27
Change time.clock() to time.process_time()

angus-g:
`time.clock()` was removed in Python 3.8 (and deprecated since Python 3.3). Use `time.process_time()` instead, which gives the sum of system and user CPU time, but doesn't include time elapsed during sleep.

I don't quite know the context for the current call: if doesn't seem to be measuring a particularly tight loop, so I don't think `time.perf_counter()` is necessary.

### DIFF
--- a/python/fluidity/regressiontest.py
+++ b/python/fluidity/regressiontest.py
@@ -133,7 +133,7 @@ class TestProblem:
         self.log("Running")
 
         run_time=0.0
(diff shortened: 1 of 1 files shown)


tmbgreaves:
Looks good to me, and we need this one in to get testharness working properly with python3.

tmbgreaves (approved):

PR #283, opened 2020-08-27
Write particle processor IDs to output

angus-g:
Because particles are uniquely identified by a (node, proc) ID pair, we need to output both these properties for diagnostics.


Cmath2 (approved):
Looks good to me!

PR #284, opened 2020-09-01
Active particle scheme

Cmath2:
This pull request aims to implement an active particle scheme within the fluidity framework. The build extends upon the now merged passive particle scheme, and enables particles to interact with, and influence the values of fields during advection. This is done in three steps:

1.  Relevant field values are interpolated onto particles at their current position in space.
2.  Particles then calculate 'attributes', based on their position, the current simulation time, and interpolated field values.
3.  Particle attributes are then interpolated back onto the nodes of a chosen mesh, with interpolated particle values able to generate new fields, or influence old ones.

With particles able to interact with and generate fields, limits must be placed on the number of particles present in set regions of the domain. This build has two primary components, the implementation of an active particle scheme able to interact with the fields of a simulation, and the implementation of spawning and deleting algorithms to control the number of particles present (per control volume). Currently, particles are spawned by cloning all parent particles that exist within a control volume and perturbing their location by a user specified amount. Conversely, particles are currently deleted by implementing a coin flip algorithm, where a coin is flipped for every particle within a control volume, and individual particles are deleted if the result is heads.

The schema has been updated to include new options for particle spawning and deleting (under each particle_group). Here the user can set the minimum and maximum thresholds of particles per control volume, as well as the location of spawned particles, either by setting a radius around the parent particle in which the new particle can spawn, or by allowing particles to spawn randomly within the current control volume prioritizing elements with fewer particles. Spawning caps can also be placed on which particle subgroups are spawned, where the ratio of particles in a specific group to the total particles in the control volume must be equal to or greater than the cap for spawning of that group to occur (e.g., if a cap of 90% is set, 90% or more of the particles in a control volume must belong to one group for spawning of that group to occur). Finally, the user can specify whether spawned particles will copy attribute values from their parents, or have their attribute values calculated as a weighted function of all particles in the control volume.

Options have also been added underneath diagnostic scalar fields, with two new algorithms, 'from_particles' and 'number_of_particles', now available.
-The 'from_particles' algorithm implements the ratio method to interpolate the given particle_group and particle_attribute values onto a user specified mesh (Currently only works for p1 meshes). Two different interpolation options are also present here, being a weighted distance interpolation of particle attributes to nodes, or nearest neighbour interpolation.
-The 'number_of_particles' algorithm simply counts the number of particles present per control volume, and sets this as the value for the relevant node. This enables users to track regions of a domain that are relatively dense, or sparse in particles.

Additional smaller changes have also been made, which include:
-With the recent h5part IO changes, the option to make particles static or write NaN outputs upon leaving a domain have been removed. Instead particles which leave a domain are simply deleted.
-Detector%types have been removed. By default all detectors are now static, and all particles are now lagrangian
-Detector%name has been removed, as they are not used in the IO, and detectors/particles are uniquely identified by the combination of their id and proc_id
-Write_diagnostics has also been updated to include the total number of particle present per particle_group, as well as each particle_subgroup within the particle_group
-It is no longer required to define the number of particles for each particle sub_group in the schema file
-Particles can now be initialised after the start of the simulation. This option appears in the schema file as initialise_during_simulation under particle_subgroup
-Profiler calls have been added to time various aspects of the particle scheme
-Several test cases have been added to test various features of the active particle scheme
-Several bug fixes to do with particle parallelization
-Updates to test case option files to account for changes to the schema


femtools/Particles.F90
```diff
@@ -856,14 +978,13 @@ subroutine create_single_particle(detector_list, xfield, position, id, proc_id,
...
     ! Populate particle
-    detector%name = name
     detector%position = position
     detector%element = element
     detector%local_coords = lcoords
-    detector%type = LAGRANGIAN_DETECTOR
     detector%id_number = id
     detector%proc_id = proc_id
-    detector_list%proc_part_count = detector_list%proc_part_count + 1
+    detector%list_id = list_id
```
angus-g, line 986: Can we get this from `detector_list%id`, and avoid having to pass the list ID through so many levels (it's a bit abstraction-breaking too)?

femtools/Particles.F90
```diff
@@ -525,33 +627,52 @@ subroutine read_particles_from_python(n_particles, subgroup_name, subgroup_path,
...
+       do i = 1, n_particles
+          call create_single_particle(p_list, list_id, xfield, coords(:,i), &
+               i+id_number, proc_num, dim, attr_counts, global=global)
+       end do
+    else
+       do i = 1, n_particles
+          call create_single_particle(p_list, list_id, xfield, coords(:,i), &
+               i, proc_num, dim, attr_counts, global=global)
+       end do
+    end if
```
angus-g, line 675: Could avoid doubling-up the loop here (and maybe making the role of `id_number` a bit clearer) with something like:

```suggestion
    offset = 0
    if (present(id_number)) offset = id_number
    do i = 1, n_particles
       call create_single_particle(p_list, list_id, xfield, coords(:,i), &
            i+offset, proc_num, dim, attr_counts, global=global)
    end do
```

femtools/Particles.F90
```diff
@@ -525,33 +627,52 @@ subroutine read_particles_from_python(n_particles, subgroup_name, subgroup_path,
...
+    call set_detectors_from_python(func, len(func), dim, current_time, coord_ptr, n_particles, stat)
+    call c_f_pointer(coord_ptr, coord_array_ptr, [dim, n_particles])
     allocate(coords(dim, n_particles))
-    call set_detector_coords_from_python(coords, n_particles, func, current_time)
+    if (n_particles==0) return
+    coords = coord_array_ptr
 
-    str_size = len_trim(int2str(n_particles))
-    fmt="(a,I"//int2str(str_size)//"."//int2str(str_size)//")"
+    call deallocate_c_array(coord_ptr)
```
angus-g, line 663: I guess this is okay after the `return` because the return is only for zero particles, in which case the pointer would be 0-length, so it's not going to leak memory...

femtools/Particles.F90
```diff
@@ -755,9 +875,10 @@ subroutine read_particles_from_file(n_particles, subgroup_name, subgroup_path, &
       call read_attrs(h5_id, dim, attr_counts%old_fields, old_field_names, old_field_vals, prefix="old%")
 
       ! don't use a global check for this particle
-      call create_single_particle(p_list, xfield, &
-           positions, id(1), proc_id(1), trim(particle_name), dim, &
-           attr_counts, attr_vals, old_attr_vals, old_field_vals, global=.false.)
+      call create_single_particle(p_list, list_id, xfield, &
+           positions, id(1), proc_id(1), dim, &
+           attr_counts, attr_vals, old_attr_vals, old_field_vals, global=.false., particle_number=particle_number)
```
angus-g, line 880: Can you use `particle_number` as the loop counter (or `i` as the particle number parameter)? Seems redundant to have both

femtools/Particles.F90
```diff
@@ -818,6 +939,8 @@ subroutine create_single_particle(detector_list, xfield, position, id, proc_id,
     !! or for the local processor only (false).
     !! This affects the inquiry of the element owning the particle
     logical, intent(in), optional :: global
+    !> Dummy counter for particle number. Removes issues when reading from checkpoint with spawning/deleting
+    integer, intent(in), optional :: particle_number
```
angus-g, line 943: Not sure I understand the point? What issues does this resolve?

femtools/Particles.F90
```diff
@@ -946,8 +1067,110 @@ subroutine move_particles(state, dt, timestep)
       call move_lagrangian_detectors(state, particle_lists(i), dt, &
            particle_lists(i)%total_attributes)
     end do
+
+    call profiler_toc("particle_advection")
   end subroutine move_particles
 
+  !> Initialise constant attribute values before diagnostic fields are set
+  subroutine initialise_constant_particle_attributes(state, subgroup_path, p_list)
```
angus-g, line 1075: It is a little bit of a shame to have to duplicate just this constant-value part out of the attribute update, but I can't really think of a better solution at the moment...

schemas/diagnostic_algorithms.rnc
```diff
@@ -100,6 +100,40 @@ internal_algorithm =
...
+	       }|
+	       ## Transfers particle values only to the closest node.
+	       element nearest_neighbour {
+	          empty
+	       }
+            }
+	 )
+      }|
+      element algorithm {
+         attribute name { "number_of_particles" },
```
angus-g, line 130: Documentation for this method?

schemas/diagnostic_algorithms.rnc
```diff
@@ -100,6 +100,40 @@ internal_algorithm =
...
          attribute name { "Internal" },
          attribute material_phase_support { "multiple" }
+      }|
+      ## This diagnostic is derived from particle attributes, it is calculated
+      ## within the Particle_Diagnostics.F90 routine, and set via the ratio method.
+      element algorithm {
+         attribute name { "from_particles" },
+	 attribute material_phase_support { "multiple" },
+	 (
+            element particle_group {
```
angus-g, line 110: Documentation for group and attribute?

angus-g (changes requested):
Awesome work! There are lots of minor changes (e.g. directly use booleans instead of comparing them, etc.)

One big thing to check is whether to use `detector_list%total_attributes` instead of trying to figure out the number of attributes off a single particle and using a collective to distribute that number. I haven't flagged all the places where this happens because it only came to me later in the review.

Another thing I wasn't so sure about is this notion of "temp" operations on particle lists -- I couldn't quite figure out why the normal versions of these aren't sufficient.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+    do i = 1,node_count(mesh)
+       if (node_part_count(i)<min_thresh) then
+          if (node_part_count(i)==0) then
+             if (node_owned(mesh, i)) then
+                call spawn_zero_particles(temp_part_count(:), node_particles(:,:), group_arrays, xfield, add_particles, i, max_thresh, copy_parents)
+                summed_particles=summed_particles+add_particles
+             end if
+          end if
+       end if
+    end do
```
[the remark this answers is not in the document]
Cmath2: In this case, we want the loop going over control volumes with 0 particles to occur after the normal spawning loop has finished. This allows the surrounding control volumes to be filled with more particles and create a more accurate spawn_zero routine

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+    proc_num = getprocno()
+
+    !Check the ratio of particles present per group and determine which groups spawn
+    spawn_group = .true.
+    if (present(cap_percent)) then
+       spawn_group = .false.
+       do i=1,size(group_arrays)
+          if ((((node_particles(i)%length*1.0)/sum(node_particles(:)%length)*1.0)*100)>=cap_percent) then
+             spawn_group = .true.
+             group_spawn = i
```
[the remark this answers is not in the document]
Cmath2: Only one group should be able to meet the cap_percent condition (we normally have it set to 99.9%), so this is intended. If the cap_percent option is on we only want one group to be spawning anyway

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+             weighted_old_attributes = weighted_old_attributes/node_particles(j)%length
+             weighted_old_fields = weighted_old_fields/node_particles(j)%length
+          end if
+       end if
+
+       !Determine id, size and shape of spawned particle parameters
+       temp_part => particle_lists(group_arrays(j))%last
+       if (associated(temp_part)) then
+
+          id = particle_lists(group_arrays(j))%proc_part_count
```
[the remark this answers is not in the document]
Cmath2: if temp_part isn't associated, we won't be using id anyway. temp_part is grabbing the last particle in the particle_list, so if it doesn't exist there won't be any particles in the list and
`particle => node_particles(j)%first`
will also be unassociated

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+          attribute_size(1) = size(temp_part%attributes)
+          attribute_size(2) = size(temp_part%old_attributes)
+          attribute_size(3) = size(temp_part%old_fields)
+          allocate(node_coord(size(temp_part%local_coords)))
```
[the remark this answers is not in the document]
Cmath2: Correct.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+          allocate(node_numbers(size(temp_part%local_coords)))
+       end if
+
+       !Spawn particles
+       particle => node_particles(j)%first
+       do while(associated(particle))
```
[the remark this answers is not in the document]
Cmath2: That's right. It duplicates every particle currently on the node.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+    logical :: delete_group
+    integer :: j
+    real :: random
+
+    !Check ratio of each particle group present and which groups will be deleted
+    delete_group = .true.
+    if (present(cap_percent)) then
+       delete_group = .false.
+       do j=1,size(group_arrays)
+          if ((((node_particles(j)%length*1.0)/sum(node_particles(:)%length)*1.0)*100)>=cap_percent) then
```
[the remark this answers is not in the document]
Cmath2: I should probably restrict deleting to this group so it's consistent with the spawning case. In reality if we have cap_percent on and the condition is met, only particles from this group should exist on the node anyway.

assemble/Sam_integration.F90
```diff
@@ -1500,8 +1500,7 @@ subroutine transfer_detectors(old_mesh, new_positions, node_ownership)
           ! Pack this node for sending
           
           ! Integer data
-          isend_data(owner)%ptr(data_index(owner) * idata_size + 1) = node%type
-          isend_data(owner)%ptr(data_index(owner) * idata_size + 2) = node%id_number
+          isend_data(owner)%ptr(data_index(owner) * idata_size + 1) = node%id_number
```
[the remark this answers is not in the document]
Cmath2: Same integration isn't used from what I've been told, so rhodri said to leave it as is.

femtools/Detector_Tools.F90
```diff
@@ -253,6 +309,27 @@ subroutine insert_into_detector_list(detector, current_list)
 
   end subroutine insert_into_detector_list
 
+  subroutine insert_into_temp_detector_list(detector, current_list)
```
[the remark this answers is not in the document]
Cmath2: The purpose of the temp_lists is to create a second 'temporary' particle lists only containing the particles within a certain particle group that exist on a specific node of the mesh (i.e. the 'node_particles' lists). Constructing these temporary lists at the beginning allows us to loop over only these particles when performing spawning/deleting instead of going through the entire particle list and checking if each particle falls within the control volume for each routine. 

The temp features are also necessary to keep the original particle lists intact when creating these new node lists and looping over them. For example, if we pick the first particle on a node and want to loop from there, we would originally have needed to use particle%next. The problem is, this will take us to the next particle on the particle_group list, not the node list. As a solution, I've added the %temp_next parameter to link particles in the node_list without interfering with how they are linked on the particle_group list.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+             call calculate_diagnostics_from_particles(state, i, s_field)
+          end if
+       end do
+       k = size(state)
+
+       !Initialise internal MaterialVolumeFraction field
+       s_field => extract_scalar_field(state(k), "MaterialVolumeFraction")
+       call calculate_sum_material_volume_fractions(state, s_field)
+       call scale(s_field, -1.0)
+       call addto(s_field, 1.0)
```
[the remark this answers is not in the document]
Cmath2: Nope, good catch!

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+        call calculate_ratio_from_particles(states, state_index, s_field)
+
+    case("number_of_particles")!Case to calculate diagnostic field from the number of particles present
+        call calculate_numbers_from_particles(s_field)
+
+    end select
+
+  end subroutine calculate_diagnostics_from_particles
+
+  subroutine calculate_ratio_from_particles(states, state_index, s_field)
```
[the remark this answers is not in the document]
Cmath2: I've combined then into a new subroutine `calculate_field_from_particles`

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+       return
+    end if
+
+    !If adjacent CV's contain particles, clone weighted particles
+    !from adjacent CV's into this CV
+    do i = 1,size(group_arrays)!Loop over particle groups
+       temp_part => particle_lists(group_arrays(i))%last
+       if (associated(temp_part)) then
+
+          id = particle_lists(group_arrays(i))%proc_part_count
```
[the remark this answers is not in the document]
Cmath2: Not necessarily, particles may exist for one particle subgroup but not another. Since we are looping over all groups here, we may come across empty ones that will not be set.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1558 @@
...
+                node_coord(4)=1-max_lcoord-rand_lcoord_2-rand_lcoord_3
+             case(4)
+                node_coord(1)=rand_lcoord_2
+                node_coord(2)=rand_lcoord_3
+                node_coord(3)=1-max_lcoord-rand_lcoord_2-rand_lcoord_3
+                node_coord(4)=max_lcoord !Will fall in this nodes CV
+             end select
+          end select
+       end if
+    end do
```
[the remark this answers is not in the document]
Cmath2: Makes sense to me!

assemble/Particle_Diagnostics.F90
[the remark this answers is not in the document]
Cmath2: I've decided to leave this as is. In the spawning case, if a group reaches the cap (e.g. 99%), we are saying we exist in that group's domain and only want to spawn particles from said group. Conversely in the deleting case, we are again saying we are in the 99% group's domain, and it would be beneficial to remove spurious particles from the other group if they exist here. Hence we delete across all groups instead of just the dominant group. Additionally, if we only deleted the dominant group here, we could be artificially reducing the ratio of the dominant group to other groups.

femtools/Detector_Tools.F90
```diff
@@ -216,6 +242,38 @@ subroutine detector_list_deallocate(detector_list)
...
+    ! Deallocate move_parameters
+    parameters => detector_list%move_parameters
+    if (associated(parameters)) then
+       if (allocated(parameters%timestep_weights)) then
+          deallocate(parameters%timestep_weights)
+       end if
+       if (allocated(parameters%stage_matrix)) then
+          deallocate(parameters%stage_matrix)
+       end if
+    end if
```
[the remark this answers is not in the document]
Cmath2: None of these parameters are ever allocated on the temp lists, so I've removed all of this.

femtools/Particles.F90
Cmath2: Seems the whole `particle_number` parameter was redundant anyway

assemble/Particle_Diagnostics.F90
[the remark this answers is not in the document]
angus-g: Gotcha. That's a subtle one, but makes sense!

assemble/Particle_Diagnostics.F90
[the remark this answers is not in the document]
angus-g: Okay, I think it was because `temp_part` comes from `particle_lists` vs. `node_particles` so I wasn't sure whether it could be unassociated but there still being something in `node_particles`. Your logic makes sense though, so it's fine!

assemble/Particle_Diagnostics.F90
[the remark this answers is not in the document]
angus-g: Aha, yep. But again this won't be an issue because if `temp_part` isn't associated, the next loop won't be doing anything useful anyway?

assemble/Particle_Diagnostics.F90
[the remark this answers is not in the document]
angus-g: Okay, this makes sense!

assemble/Sam_integration.F90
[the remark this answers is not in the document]
angus-g: No problem, so this is just a "make it compile" change?

femtools/Detector_Tools.F90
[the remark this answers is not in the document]
angus-g: Great, that helps a lot! I guess I was missing the node specificity requirement.

assemble/Particle_Diagnostics.F90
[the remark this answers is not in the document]
angus-g: Nice! I hope it wasn't too awkward, but looks fine

angus-g (commented):
Awesome work @Cmath2! I'll avoid giving an explicit "approve" review, but consider this my approval and hand over to @drhodrid.

angus-g (approved):
Never mind, GitHub will stick on "changes requested" unless I say otherwise!

assemble/Particle_Diagnostics.F90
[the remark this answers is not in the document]
Cmath2: Yes that's true. Now we aren't getting attribute sizes from them the if condition is redundant.

drhodrid:
@Cmath2 - since @angus-g has looked through the code carefully and @stephankramer will doubtless do so as well, following my review, I looked at the other aspects. In general, I think you've done a great job of ticking all boxes! The only things missing in my opinion are:

1. Manual - the manual needs to be updated to highlight what can now be done with "active" particles and the various underlying assumptions/methods documented. In addition, .flml options need documentation in the manual.
2. Given that this is some blooming awesome new Fluidity functionality, I would recommend adding an example highlighting a commonly used case (under examples). I'd suggest the mu10 RT case for this as it's a little bit awesome. See examples/stokes_square_convection for a suggested structure to follow.
3. At least 3 of your tests are too long for the short or medium test suite. These are particle_stable_layer, particle_unstable_layer and particle_vortex_3d. These should be moved across to longtests post-merge (merely flagging this now so it's recorded). If there are others that I'm missing, please remember to move those across as well. 
4. The comments in the .rnc/rng are fine in general, but a few could do with updating (e.g. text that comes up when you activate "particle_group".
5. Test case .xmls - I'd recommend adding a little more information on what is being tested and how it is being tested to the test descriptions. This will ensure that years down the line, if specific tests start to fail we have a better idea of where to look for the problem. 
6. Test case names - perhaps be a little more specific for some? E.g. Particle_multimaterial --> particle_rayleigh_taylor_instability (or similar). Others (e.g. particle_spawning) are as clear as they can be. 

Aside from that, you've done blooming awesome. Now over to @stephankramer for the more careful checks!

drhodrid:
Thanks for doing such a thorough job addressing my comments @Cmath2. I guess we now hand over to @stephankramer for the ultimate test!!

stephankramer:
Thanks, I had indeed waited for @angus-g - who with @drhodrid, seem to have done a pretty thorough job already - let's see if I can still find something to moan about..

Cmath2:
Ok @stephankramer I believe the branch is ready for you again. We had a couple of bugs arise regarding attribute_arrays which have been handled, and I've also added another test case to ensure they work.

assemble/Sam_integration.F90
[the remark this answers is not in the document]
stephankramer: Yes, sam_integration is what we used before zoltan, and always only supported a subset of discretisations (no DG). It was kept around for a long time for people that don't have zoltan, but is completely untested afaics so I just opened a ticket to remove it: #288 - if someone's feeling virtuous ;-)

assemble/Zoltan_integration.F90
```diff
@@ -2073,17 +2072,10 @@ subroutine update_particle_list_element(detector_list_array)
        detector_list => detector_list_array(j)%ptr
        ewrite(2,*) "Length of detector list to be updated: ", detector_list%length
 
+       if (sum(detector_list%total_attributes)==0) return
```
stephankramer, line 2075: what's going on here? Can we just ignore any other lists in detector_list_array? This also assumes that `detector_list%total_attributes` is the same on all processes - which seems to contradict what's happening further below in the new bcasts...

assemble/Zoltan_integration.F90
```diff
@@ -2130,6 +2122,15 @@ subroutine update_particle_list_element(detector_list_array)
           deallocate(ndets_being_sent)
           cycle
        end if
+       !Check if any are being sent
+       do i = 1, size(ndets_being_sent)
+          if (ndets_being_sent(i)>0) then
+             call mpi_bcast(detector_list%total_attributes, 3, getPINTEGER(), i-1, MPI_COMM_FEMTOOLS, ierr)
```
stephankramer, line 2128: Why do we need to communicate this? Do other processes just get their `%total_attributes` overwritten?

assemble/Zoltan_integration.F90
```diff
@@ -2130,6 +2122,15 @@ subroutine update_particle_list_element(detector_list_array)
...
+             assert(ierr == MPI_SUCCESS)
+             total_attributes=sum(detector_list%total_attributes)
+             exit
```
stephankramer, line 2131: What happens if there are more than one processes that want to send?

assemble/Zoltan_integration.F90
```diff
@@ -2059,8 +2059,7 @@ subroutine update_particle_list_element(detector_list_array)
     integer :: old_local_element_number, new_local_element_number, old_universal_element_number
     integer, allocatable :: ndets_being_sent(:)
     real, allocatable :: send_buff(:,:), recv_buff(:,:)
-    logical do_broadcast
-    integer, dimension(3) :: attribute_size
+    logical :: do_broadcast, sent
```
stephankramer, line 2062: I don't think `sent` is used

assemble/Zoltan_integration.F90
Cmath2: Good catch. This should cycle to the next detector list if `sum(detector_list%total_attributes)==0`. 
`detector_list%total_attributes` will be the same on all processes, I'll fix up the relevant contradictions below.

assemble/Zoltan_integration.F90
Cmath2: This is a relic from when attribute_size was being sent. Back then it was needed to give the size to processors which may have not contained particles beforehand, but with the size stored on detector_lists this is no longer necessary.

assemble/Zoltan_integration.F90
Cmath2: This whole loop is now redundant. It was previously checking if any processors were sending, and if they were it would set the value for total_attributes from the passed in attribute_size value from that processor.

femtools/Detector_Tools.F90
```diff
@@ -335,14 +426,26 @@ subroutine delete_all_detectors(detector_list)
 
   end subroutine delete_all_detectors
 
+  subroutine delete_all_temp_detectors(detector_list)
```
stephankramer, line 429: Again I wouldn't call it `delete_all_temp_detectors`as it does not delete or deallocate detectors.

I would just merge this with `temp_deallocate` above (and call it `temp_list_deallocate`). Strictly speaking it doesn't do any deallocation, it just wipes the temp links. Even that nullifying of the temp links in principle isn't necessary, as they'll simply be overwritten the next time a detector is inserted into a temp list. So in principle the user shouldn't have to do anything when it's done with the temp list, but it is probably neater to do this clean up (and allows to later change the implementation where some deallocation might be necessary).

femtools/Detector_Tools.F90
```diff
@@ -600,7 +699,7 @@ subroutine set_detector_coords_from_python(values, ndete, func, time)
     !!< specified in the string func at those points. 
     real, dimension(:,:), target, intent(inout) :: values
     !! Func may contain any python at all but the following function must
-    !! be defiled:
+    !! be defined:
```
stephankramer, line 702: Nice one! :-)

femtools/Halos_Communications.F90
```diff
@@ -565,6 +569,101 @@ subroutine halo_update_tensor(t_field, level, verbose)
     end if
     
   end subroutine halo_update_tensor
+
+  subroutine halo_accumulate_array_real_star(halo, real_data, block_size)
+    !!< For nodes that are seen by one or more other processors (recv nodes on these processes)
+    !!< add the contributions they have stored for these recv nodes to the correpsonding entry
```
stephankramer, line 575: *corresponding*

main/Fluids.F90
```diff
@@ -228,6 +231,9 @@ SUBROUTINE FLUIDS()
          & default=1)
     call get_option("/timestepping/nonlinear_iterations/tolerance", &
          & nonlinear_iteration_tolerance, default=0.0)
+
+    !Initialise positions of particles before adapt_at_first_timestep
```
stephankramer, line 235: Sorry to be pedantic: but could you add a space after the comment marker (and in the lines below)

schemas/diagnostic_algorithms.rnc
```diff
@@ -100,6 +100,55 @@ internal_algorithm =
       element algorithm {
          attribute name { "Internal" },
          attribute material_phase_support { "multiple" }
+      }|
+      ## This diagnostic is derived from particle attributes, it is calculated
+      ## within the Particle_Diagnostics.F90 routine, and set via the ratio method.
+      element algorithm {
+         attribute name { "from_particles" },
+	 attribute material_phase_support { "multiple" },
```
stephankramer, line 108: The indentation seems off here

schemas/diagnostic_algorithms.rnc
```diff
@@ -100,6 +100,55 @@ internal_algorithm =
...
+	 (
+	    ## Particle group used for calculation of diagnostic field.
```
stephankramer, line 110: and here

schemas/fluidity_options.rnc
```diff
@@ -276,17 +266,18 @@ start =
                }?
             }?
          },
-         ## Particle options. Particles, like Lagrangian detectors, move with a specified velocity field.
+         ## Particle options. Lagrangian particles which move with a specified velocity field.
```
stephankramer, line 269: Options for particles, Langrangian particles which ....

schemas/fluidity_options.rnc
```diff
@@ -276,17 +266,18 @@ start =
...
          ## In addition, particles can carry attributes, being properties of the particles. These attributes
          ## can be constant, or based on a python expression that depends on other attribute values, at the end
-         ## of the previous timestep, or field values interpolated at the particle location. Planned future
-         ## functionality includes the ability to spawn and delete particles to ensure sufficient particles are
-         ## present in every region, and the ability to interpolate from particles back to the (Eulerian) mesh.
+         ## of the previous timestep, or field values interpolated at the particle location. Attributes values
+	 ## can be interpolated back onto the (Eulerian) mesh through the diagnostic algorithm from_particles.
```
stephankramer, line 273: Indentation

schemas/fluidity_options.rnc
```diff
@@ -276,17 +266,18 @@ start =
...
+	 ## Options can also be set to spawn and delete particles when the number of particles within a control
+	 ## volume either exceeds or falls below a specified threshold value.
          element particles {
             (
-               ## Particles groups containing multiple particle arrays or subgroups. The particle group specifies
-               ## which velocity field is used, as well as attributes carried and (eventually) spawning
-               ## and deleting options.
+               ## Particle groups containing multiple particle arrays or subgroups. The particle group specifies
+               ## particle IO options, as well as common attributes carried per particle subgroup for diagnostic
+	       ## calculations and spawning and deleting options.
```
stephankramer, line 280: Indentation

schemas/fluidity_options.rnc
```diff
@@ -335,20 +326,99 @@ start =
                       element wall_time_dump_period {
                           real
                       }?
+		  }?,
+                  ## Options for spawning and deleting particles based on the minimum and maximum limits of
```
stephankramer, line 330: Indentation

schemas/fluidity_options.rnc
```diff
@@ -335,20 +326,99 @@ start =
...
+		        element percentage {
+			   real
+		        }
+		     }?
                   }?,
-                  ## Particle subgroups arrays specified within a particle group. All subgroups utilize the same velocity field
-                  ## and carry the same attributes. Subgroups can have different initial positions specified via a python
-                  ## function of from a file.
+                  ## Particle subgroup arrays specified within a particle group. Particle attribute values are specified per
+		  ## subgroup, although all subgroups should be set to carry the same attribute names if it is desired to calculate
```
stephankramer, line 386: Indentation

schemas/fluidity_options.rnc
```diff
@@ -335,20 +326,99 @@ start =
...
                      attribute name { xsd:string },
-                     ## The number of particles prescribed by the python function.
-                     element number_of_particles {
-                        integer
-                     },
                      (
                         ##Initial position of particles
                         element initial_position {
-                           ## Python function prescribing dimensional vector input. Functions should be of the form:
+			   (
```
stephankramer, line 394: I think you don't need the round brackets here - in any case indentation is off

schemas/fluidity_options.rnc
```diff
@@ -335,20 +326,99 @@ start =
...
+                              ##
+                              ## *** IMPORTANT NOTE ***
+                              ##
+                              ## The t argument is not used here. To spawn with t dependence use initialise_during_simulation option below.
+                              element python {
+                                 python_code
+                              }|
+                              ## File containing the particle positions in binary form
+                              element from_file {
+			         (
```
stephankramer, line 411: again the round brackets are spurious

schemas/fluidity_options.rnc
```diff
@@ -358,15 +428,11 @@ start =
                            ##
                            ## *** IMPORTANT NOTE ***
                            ##
-                           ## The t argument is for future use only - currently particle locations are only set at simulation start.
+                           ## This function will only be called after simulation start. To initialise starting positions use initial_position option above.
```
stephankramer, line 431: use *the* initial_position option above

schemas/fluidity_options.rnc
```diff
@@ -389,9 +455,9 @@ start =
 				 },
 				 ## Do not output this attribute to file
 				 element exclude_from_output {
-				     empty
+				    empty
 				 }?,
-                                 generic_scalar_particle_attribute_choice
+                                 generic_scalar_particle_attribute_array
```
stephankramer, line 460: Indentation

schemas/fluidity_options.rnc
```diff
@@ -412,7 +478,7 @@ start =
 				 element exclude_from_output {
 				     empty
 				 }?,
-                                 generic_vector_particle_attribute_choice
+                                 generic_vector_particle_attribute_array
```
stephankramer, line 481: Guess what?

schemas/fluidity_options.rnc
```diff
@@ -433,24 +499,21 @@ start =
 				 element exclude_from_output {
 				     empty
 				 }?,
-                                 generic_tensor_particle_attribute_choice
+                                 generic_tensor_particle_attribute_array
```
stephankramer, line 502: and here

schemas/fluidity_options.rnc
```diff
@@ -433,24 +499,21 @@ start =
...
             )*,
-            (
-               ## By default Fluidity will fail if a particle has left the domain.
-               element fail_outside_domain {
-                  empty
-               }|
-               ## Enable to write NaN values to particle output if a particle has left the domain.
-               element write_nan_outside_domain {
+            ## Options for particles being advected outside the computational domain.
+	    element moving_outside_domain {
```
stephankramer, line 511: indentation

stephankramer (changes requested):
Wow! This is a lot of hard work - and I still haven't gotten through everything!

Nothing serious so far: for the temp_list stuff, just add/correct some comments and rename some of the routines.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1420 @@
...
+    !Initialise constant particle attributes
+    list_counter=1
+    do i = 1,particle_groups
+       group_path = "/particles/particle_group["//int2str(i-1)//"]"
+       do j = 1,particle_arrays(i)
+          particle => particle_lists(list_counter)%first
+          subgroup_path = trim(group_path) // "/particle_subgroup["//int2str(j-1)//"]"
+          if (option_count(trim(subgroup_path) // "/attributes/scalar_attribute/constant") + &
+               & option_count(trim(subgroup_path) // "/attributes/vector_attribute/constant") + &
+               & option_count(trim(subgroup_path) // "/attributes/tensor_attribute/constant").gt.0) then
```
stephankramer, line 103: Nitpicking on style: could you just use `>`, `<`, `>=`, `<=` and `==` rather than `.gt.', etc.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1420 @@
...
+          call get_particle_arrays(lgroup, group_arrays, group_attribute, att_n=att_n, lattribute=lattribute)
+       else
+          call get_particle_arrays(lgroup, group_arrays)
+       end if
+       
+
+       !Allocate arrays to store summed attribute values and particle counts at nodes
+       if (is_ratio) then
+          allocate(node_values(node_count(s_field)))
+          node_values(:) = 0
```
stephankramer, line 383: Style: just say `node_values = 0`

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1420 @@
...
+       end if
+       allocate(node_part_count(node_count(s_field)))
+       node_part_count(:) = 0
+
+       if (is_ratio .and. have_option(trim(complete_field_path(s_field%option_path))// "/algorithm/interpolation/weighted_distance")) then
+
+          !Loop over particle arrays
+          do i = 1,size(group_arrays)
+             particle => particle_lists(group_arrays(i))%first
+             if (associated(particle)) then !Only work on arrays if local_particles exist on this processor
```
stephankramer, line 393: You could replace this with `if (.not. associated(particle)) cycle`
Then on the next line you always allocate `local_crds` (if it gets there), and you don't have to check `if (allocated(local_crds))` after the particle loop.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1420 @@
...
+                !Increase particle count for this node by 1
+                node_part_count(node_number) = node_part_count(node_number) + 1.0
+                particle => particle%next
+
+             end do
+             if (allocated(local_crds)) then
+                deallocate(local_crds)
+             end if
+          end do
+       end if
```
stephankramer, line 451: Does this `if-block` need an else? It's a little unclear what the cases are; Maybe add a comment?

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1420 @@
...
+             if (is_ratio) then
+                !Determine field value from ratio method
+                ratio_val = node_values(i)/node_part_count(i)
+                s_field%val(i) = ratio_val
+             else
+                !Store number of particles on field
+                s_field%val(i) = node_part_count(i)
+             end if
+          end if
+       end do
```
stephankramer, line 472: This is probably a little faster:
```
if (is_ratio) then
  where (node_part_count/=0)
      s_field%val = node_values/node_part_count
  end where
else
  s_field%val = node_part_count
end if
```

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1420 @@
...
+
+                !Find nodes for specified element
+                nodes => ele_nodes(s_field, element)
+
+                !Distribute particle values across nodes of element by weighted distance function
+                do j = 1, size(nodes)
+                   node_number = nodes(j)
+                   node_values(node_number) = node_values(node_number) + att_value*local_crds(j)
+                   node_part_count(node_number) = node_part_count(node_number) + 1.0*local_crds(j)
+                end do
```
stephankramer, line 410: Could write:
```
node_values(nodes) = node_values(nodes) + att_value*local_crds
node_part_count(nodes) = node_part_count(nodes) + local_crds
```

stephankramer (changes requested):
Some more feedback on Particle_Diagnostics.F90

femtools/Detector_Parallel.F90
```diff
@@ -506,4 +512,39 @@ subroutine sync_detector_coordinates(state)
...
+    allocate(proc_set(getnprocs()))
+    call mpi_allgather(val, 1, getPINTEGER(), proc_set, 1, getPINTEGER(), MPI_COMM_FEMTOOLS, ierr)
+    assert(ierr == MPI_SUCCESS)
+    do i = 1, size(proc_set)
+       if (proc_set(i)==1) then
+          call mpi_bcast(detector_list%total_attributes, 3, getPINTEGER(), i-1, MPI_COMM_FEMTOOLS, ierr)
+          assert(ierr == MPI_SUCCESS)
+          exit
+       end if
+    end do
```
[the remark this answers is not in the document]
Cmath2: I don't think this routine is needed at all anymore. All processors should have values assigned to %total_attributes from Particles.F90, so this routine can simply be replaced with distribute_detectors

Cmath2:
I believe everything is up to date and ready for the next round @stephankramer

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1347 @@
...
+       ! target node number
+       if (node_num /= node_numbers(i)) cycle
+
+       ! set the coordinates according to the permutation for this index
+       do j = 1, size(node_coord)
+          node_coord(j) = work(permutation(j,i))
+       end do
+
+    end do
+
```
stephankramer, line 1213: This is already complicated by the permutation, but maybe you can simplify a bit, showing that this is really just two loops consecutively:
```
  do i=1, size(node_coord)
    if (node_num == node_numbers(i)) exit
  end do
  assert(i<=size(node_coord)
  do j=1, size(node_coord)
      node_coord(j) = work(permutation(j,i))
  end do
```

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1347 @@
...
+  end subroutine set_spawned_lcoords
+
+  subroutine local_to_global(coordinates, l_coords, ele_number, global_coord)
```
stephankramer, line 1216: Why is this subroutine needed? I think you can just do `global_coord = eval_field(ele_number, coordinates, l_coords)`. If it's just to normalize the l_coords, you can just do `global_coord = eval_field(ele_number, coordinates, l_coords/sum(l_coords))` where needed.

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1347 @@
...
+    logical :: delete_group
+    integer :: j
+    real :: rand_val
+
+    !Check ratio of each particle group present and which groups will be deleted
+    delete_group = .true.
+    if (present(cap_percent)) then
+       delete_group = .false.
+       do j=1,size(group_arrays)
+          if ((((node_particles(j)%length*1.0)/sum(node_particles(:)%length)*1.0)*100)>=cap_percent) then
```
stephankramer, line 1310: Same comment as in spawn_particles()

schemas/diagnostic_algorithms.rnc
```diff
@@ -100,6 +100,55 @@ internal_algorithm =
...
+	       (
+	          ## Particle attribute used for calculation of diagnostic field.
+	          element particle_attribute {
+	             attribute name { xsd:string }
+	          }|
+	          ## Particle attribute_array used for calculation of diagnostic field.
+	          element particle_attribute_array {
+	             attribute name { xsd:string },
+		     ## Integer value corresponding to attribute position in array (starts from 1).
+		     element integer_val {
```
stephankramer, line 122: Should be `integer_value` to be consistent with rest of the schema

tests/particle_stratified_stable_layer/setup_vals.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
```
stephankramer, line 1: ~~We use `#!/usr/bin/env python3` everywhere~~
This is just a python *module* (i.e. only imported) so no need for a header (should have been python3 anyway).

tests/particle_stratified_unstable_layer/setup_vals.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
```
stephankramer, line 1: This is just a python *module* (i.e. only imported) so no need for a header (should have been python3 anyway).

examples/particle_rayleigh_taylor_mu10/RTplot.py
```diff
@@ -0,0 +1,74 @@
+# This sript plots up the RMS velocity:
```
stephankramer, line 1: Let's add a `#!/usr/bin/env python3` to be consistent

stephankramer: Also, can you do a `chmod +x  examples/particle_rayleigh_taylor_mu10/RTplot.py` (and then commit), this makes the script executable, so users can just do `./RTplot.py` if they're in the directory.

femtools/Detector_Data_Types.F90
```diff
@@ -111,6 +105,12 @@ module detector_data_types
      !! Pointers for detector linked lists
      TYPE (detector_type), POINTER :: next=> null()
      TYPE (detector_type), POINTER :: previous=> null()
+     !! Pointers to temporary linked lists used during spawning and deleting.
+     !! These lists are used to form temporary linked lists within Particle_Diagnostics.F90
+     !! Temporary linked lists are created per control volume to allow for easy looping
+     !! over particle within that control volume during spawning and deleting.
+     TYPE (detector_type), POINTER :: temp_next => null()
+     TYPE (detector_type), POINTER :: temp_previous => null()
```
stephankramer, line 113: Thanks for the comments! Can you also fix the SHOUTING? :-)

tests/particle_entrainment_of_dense_layer/setup_vals.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
```
stephankramer, line 1: This is just a python *module* (i.e. only imported) so no need for a header (should have been python3 anyway).

stephankramer (changes requested):
Woohoo! Last batch of comments for me

 :1st_place_medal: 

Once again, an incredible amount of work has been done here. Thanks for patiently going through my corrections (just a few more now!). Also thanks for adding extensive tests and documentation.

Some last feedback: I have a feeling that some further optimisation can be done in the future in the spawning and deletion - I don't know if you've ever done any profiling with that? Just curious if you have, no need to do now, as I said *potential future* optimisation

I'll wait till you've pushed the final (minor) corrections to hit approve and then we can merge! Good luck with the final hurdles of your PhD!

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1347 @@
...
+       if (is_ratio) then
+          call halo_accumulate(halo, node_values)
+       end if
+    end if
+
+    if (is_ratio) then
+       !Store value on field (if node has at least one particle)
+       where (node_part_count/=0)
+          !Determine field value from ratio method
+          s_field%val = node_values/node_part_count
```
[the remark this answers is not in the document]
Cmath2: s_field%val was previously set to 0 on line 285, but this location probably suits better

assemble/Particle_Diagnostics.F90
```diff
@@ -0,0 +1,1347 @@
...
+          s_field => extract_scalar_field(state(i),k)
+          if (have_option(trim(s_field%option_path)//"/diagnostic/algorithm::from_particles")) then
+             call get_option(trim(s_field%option_path)//"/name", name)
+             if (name=="MaterialVolumeFraction") cycle
+             call calculate_field_from_particles(state, i, s_field)
+          else if (have_option(trim(s_field%option_path)//"/diagnostic/algorithm::number_of_particles")) then
+             call calculate_field_from_particles(state, i, s_field)
+          end if
+       end do
+    end do
```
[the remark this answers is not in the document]
Cmath2: Whoops, this subroutine was meant to be deleted

Cmath2:
Hey @stephankramer, I believe I've addressed everything except for the random_seed call.

For this the goal is to create a random seed so there is consistency in the random numbers generated for spawning/deleting between runs. I do see the point that calling random_seed every time particle_cv_check is called is redundant though. If we are to keep the seed, would you suggest I instead set it within Fluids.F90 before the first call to particle_cv_check (e.g. on line 342)? Or somewhere else like in initialise_particles with a check that the spawning/deleting option is active?

Regarding your other feedback, I don't believe I've done any thorough profiling of the current spawning/deleting scheme. It probably would be best to leave this one for Angus in a future commit.

schemas/diagnostic_algorithms.rnc
Cmath2: integer_value appears to only be used for cases with a set selection of options (e.g. when selecting dim). I've changed it to attribute_int instead for clarity that it's a user set parameter.

stephankramer: Do you mean it didn't work when you tried (because diamond or spud treats `integer_value` specially?). Sorry to be pedantic on this but `attribute_int` isn't a good alternative. The main reason I objected to `integer_val` was that it abbreviates value - and we have tried very hard to avoid these in the schema. If `integer_value` doesn't work, maybe `attribute_index` ?

Also (final request I promise :crossed_fingers: ), I see there's tabs in various places you've edited the schema - could you please change these to spaces (use an editor that shows you tabs vs. spaces). Tabs only work if you use them consistently everywhere.

stephankramer:
Yeah there's probably some bike-shedding we could do about the random seed. One of the potential issues is that it affects other places in the code that uses random numbers - and in some cases you want deterministic behaviour and sometimes you don't (turbulence modelling). Probably the general solution is to have a central place to do this, a user choice whether things are deterministic or not, and also a mechanism to checkpoint the seed value. At the moment I think this only affects synthetic bcs with LES turbulence modelling, which nobody uses (in particular not in combination with particles!) - so let's just ignore this for now.

Cmath2:
I believe that covers all of the tabs in the schemas @stephankramer, let me know if I missed any. With that wrapped up I think we should be good to merge when the light goes green :)

stephankramer (approved):
Amazing - thanks for your patience with ironing out the details...
I'll leave it to you to hit that green button then :-)

PR #286, opened 2020-09-23
Remove Python 2 from python_validate test

angus-g:
Since we no longer expect to maintain backwards compatibility, this test is generating unnecessary warnings.


drhodrid (approved):
Looks good to me (pending a green Jenkins, and hopefully the disappearance of the long lived python validation warning).

PR #293, opened 2020-11-27
Fix Gtk Widget methods deprecation warning in statplot

Patol75:
Two methods have been renamed, I have updated the code accordingly.
Feel free to approve, anyone. :)

### DIFF
--- a/tools/statplot.py
+++ b/tools/statplot.py
@@ -93,8 +93,8 @@ class StatplotWindow(Gtk.Window):
     def CreateCombo(store):
         comboBox = Gtk.ComboBox.new_with_model_and_entry(store)
         comboBox.set_entry_text_column(0)
-        comboBox.set_margin_left(100)
-        comboBox.set_margin_right(100)
+        comboBox.set_margin_start(100)
+        comboBox.set_margin_end(100)
         comboBox.set_wrap_width(3)
         return comboBox
 


stephankramer:
Looks good, thanks! One query: are the new names also available with older gtk versions on slightly older installs that we might care about?

Note that we're currently experiencing some Jenkins issues (github's not talking to Jenkins) which I believe @tmbgreaves  is looking into...

Patol75:
That is indeed a good question. I have had a look and the doc says that these "new" methods have been introduced with version 3.12, which seems to date back to 2014, which is quite a surprise to me because yesterday was the first day I ran into this deprecation warning. Additionally, I could find mentions of these methods in some GitHub threads in July 2018, which again indicates that we should be pretty safe in terms of backwards compatibility.

tmbgreaves:
Jenkins should be fixed now - sorry for the problems. Short version - we renewed our SSL certificate and I got the certificate chain in the wrong order when reconstructing the keystore...

Patol75:
@tmbgreaves No worries at all, that must have been a nightmare to figure out...

stephankramer (approved):
Looks good, maybe change the `gi.require_version` on top?

Patol75:
What should I change them to?

stephankramer:
Ah I meant - if we indeed depend on 3.12 now - should it read `require_version('Gtk', '3.12')` ? I don't know, I've never done any gtk stuff

Patol75:
I see, well I have tried it, and it does not seem to be a thing actually. So I will just merge as is for now.

PR #296, opened 2020-12-26
Fix for Python3 time in regressiontest

tmbgreaves:
time.clock() no longer exists in python3 - we need to use time.perf_counter() instead. I'm not sure why this worked up to now - I'd have expected this to fail when we did the grand purge of python2.


angus-g:
Probably supersedes #282

tmbgreaves:
Thanks @angus-g , I'd completely missed that one - my bad! I'll approve and merge yours.

PR #298, opened 2021-01-02
Fix compilation issues with PETSc 3.13+

### LINKED ISSUE #292: Error while compiling Fluidity with PETSc 3.14
Hi,
This is just a heads-up that `make` fails with the newer versions of PETSc (starting somewhere around 3.13.X I believe) - I paste the errors I am getting below. I, sadly, do not quite understand what these errors mean and how to fix them. If anyone more knowledgeable has some time to investigate, please feel free to open a PR. Thank you for your time.

```
Solvers.F90:1930:89:

 1930 |         call PCBJACOBIGetSubKSP(pc, PETSC_NULL_INTEGER, PETSC_NULL_INTEGER, subksp, ierr)
      |                                                                                         1
Error: There is no specific subroutine for the generic ‘pcbjacobigetsubksp’ at (1)
Solvers.F90:1932:85:

 1932 |         call PCASMGetSubKSP(pc, PETSC_NULL_INTEGER, PETSC_NULL_INTEGER, subksp, ierr)
      |                                                                                     1
Error: There is no specific subroutine for the generic ‘pcasmgetsubksp’ at (1)
Solvers.F90:1955:88:

 1955 |        call PCBJACOBIGetSubKSP(pc, PETSC_NULL_INTEGER, PETSC_NULL_INTEGER, subksp, ierr)
      |                                                                                        1
Error: There is no specific subroutine for the generic ‘pcbjacobigetsubksp’ at (1)
```

gnikit:
This simply changes the output variables from NULLs to Fortran integers because gfortran was complaining about the types of the arguments passed to store the output, we don't actually use these output variables anywhere, so this should change absolutely nothing result or performance wise.

Fixes #292

### DIFF
--- a/femtools/Solvers.F90
+++ b/femtools/Solvers.F90
@@ -1865,6 +1865,7 @@ subroutine create_ksp_from_options(ksp, mat, pmat, solver_option_path, parallel,
     PCType:: pctype, hypretype
     MatSolverType:: matsolvertype
     PetscErrorCode:: ierr
+    integer :: n_local, first_local
     
     call get_option(trim(option_path)//'/name', pctype)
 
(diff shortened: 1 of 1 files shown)


stephankramer (approved):
Perfect, thanks for that!

Issue #289, opened 2020-10-07
Division by zero in Write_State.F90


gnikit:
In line 91
https://github.com/FluidityProject/fluidity/blob/2ad88a413ca999d4bd37b50e8ddb2588866335c1/femtools/Write_State.F90#L91

A comparison for `int_dump_period == 0` is done followed by a `mod(,int_dump_period)`, `mod()` however is incapable of handling divisions by zero (it should throw and arithmetic exception if divided by zero). The actual problem arises due to the fact that the Fortran standard does not evaluate logic statements in any particular direction and does not have short-circuit evaluation either.

>7.1.5.4.2 Evaluation of logical intrinsic operations
Once the interpretation of a logical intrinsic operation is established, the processor may evaluate any other
expression that is logically equivalent, provided that the integrity of parentheses in any expression is not violated.

taken from [this discussion](https://community.intel.com/t5/Intel-Fortran-Compiler/Evaluation-of-if-statements/td-p/879535) at the Intel Forum

So it is up to the compiler and its optimiser to chose how such clauses are evaluated. I think GCC strictly follows the standard in newer versions (at least with debugging symbols) so it will probably segfault in Ubuntu 20.04

stephankramer:
That's right: fortran does not guarantee short-circuit logic

Maybe something like:
```
if (int_dump_period>0) then
   zero_modulus = mod(timestep, int_dump_period) == 0
else
   zero_modulus = True
end if
if (zero_modulus) then
   ...
```
TBH I find the logic in that part of the code a little hard to follow

gnikit:
Closed via #301

PR #301, opened 2021-02-06
Fix short-circuit evaluation division by zero GCC v9+

gnikit:
Fixes Division by zero in Write_State.F90 #289

Fortran does not have short-circuit evaluation!

NOTE: I am not sure I have removed all the short-circuit expressions.
I do believe however that I have removed all the divs by 0 in mod.
Validated using Rust flavoured regex: `mod\((.*)\)(\s*?)`

-------

This should probably be merged to master before the focal update or merged into the focal update PR

### DIFF
--- a/assemble/qmesh.F90
+++ b/assemble/qmesh.F90
@@ -69,7 +69,7 @@ contains
     real, intent(in) :: current_time
     integer, intent(in) :: timestep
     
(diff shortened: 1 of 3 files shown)


stephankramer (approved):
Perfect - thanks for that

PR #303, opened 2021-02-20
VTK Quads and Hexas output bug

### LINKED ISSUE #302: VTK mesh output for quads and hexahedrals is inconsistent
For some strange reason the renumbering of node connectivity for VTK quads and hexahedrals is separated between VKT_interfaces.F90, vtkfortran.cpp and vtkmeshio.cpp. 

For example this results into the following bug if one decides to create a quad mesh and output it into a VTU

image

while the output mesh should be looking like this

image

I already have a fix for this and I will open a PR soon, but in general I think that libvtkfortran needs a revision and a refactoring to modernise the code (maybe a job for another time). I will make sure I add a unittest testing as many VTK elements I can.

gnikit:
Fixes #302 VTK output

Quads and hexahedrals are now drawn correctly when calling VTK.

A unit test has been added testing our 4 main elements
triangles, quads, tets and hexas. The test checks that the
DG DOFs get correctly reordered from/to Fluidity to/from VTK.


gnikit:
The `test_inverse` unittest is misbehaving again, we might have to drive up the tolerance on that one.

gnikit:
I think I might also have to delete/change the Python function under `vtutools.py:ToVtkNodeOrder`, since quad ordering is handled internally in Fortran

stephankramer (approved):
Looks good to me. As you have probably noticed the support for quads and hexes is very limited in (main?) fluidity anyway. I would indeed just remove these routines in vtutools.py (and the places that call it). This might affect checkpointing (i.e. the only time we read in vtu files), but I'm not sure that even works with quads/hexes - you should probably try to see if it does, and if so if it correctly reads things in.

gnikit:
> just remove these routines in vtutools.py (and the places that call it). 

As far as I could tell there are no places that make use of the vtutools.py code for quads, so all should be good, but I will have another look to be safe.

> This might affect checkpointing (i.e. the only time we read in vtu files), but I'm not sure that even works with quads/hexes - you should probably try to see if it does, and if so if it correctly reads things in.

I will have a look, but from what I remembered all looked good on my end. Will push shortly.

gnikit:
Apologies this took so long. I think this should be good to go now @stephankramer.
I also rechecked the chekpoint output and restart using the `rename_checkpoint` test and the `gmsh -2 -algo delquad src/mesh2.geo` to get a mixed triangular/quad mesh, then restarted a checkpoint and it all seems good, which makes sense since we simply moved to one location all the node mappings from Fluidity to VTK elements and nothing more.

The Python diagnostic node ordering methods have been deleted since they do nothing other than returning a copy of the nodes array.

gnikit:
Should this be merged or do we have to wait for the particle tests to be fixed in master?

tmbgreaves:
Rerunning tests via Actions - @stephankramer assuming no critical bugs come out of that are you happy for this to be merged? Keen to do so before master->main branch renaming on 26th if possible to avoid the mess of retargetting the PR.

tmbgreaves:
OK, we do have an interesting bug, which is that PRs from outside the Fluidity repo don't have the requisite auth details to do the Actions build and push/pull from docker.

stephankramer (approved):
Yes, happy for this to go in once green.

Re. the issue of PRs from other repos: it'd be nice if there is a solution, but I think it would be perfectly reasonable and manageable to request people to push their branch to the FluidityProject fluidity repo - which would require adding the contributor to the fluidity team. In the case of @gnikit that should of course have happened already - just added him now...

PR #306, opened 2021-03-17
Add PETSC_DIR mention to config help

adigitoleo:
This adds mention of the `PETSC_DIR` environment variable to the CLI help of the configure script. I thought it would be nice to add since PETSc is a fairly critical dependency of fluidity.

While I was at it I sorted the env vars alphabetically.

### DIFF
--- a/configure
+++ b/configure
@@ -1511,30 +1511,31 @@ Optional Packages:
                           specify the path to libnuma, e.g. -L/usr/lib -lnuma
 
 Some influential environment variables:
+  CC          C compiler command
+  CFLAGS      C compiler flags
+  CPP         C preprocessor
+  CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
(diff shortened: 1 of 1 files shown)


gnikit:
So the configure script is autogenerated using GNU autotools, that is why some entries in the configure are out of order.
If you wish to make any permanent changes to the configure script you need to edit the `configure.in`.
Now whether or not this is a useful change for master I will leave it for the maintainers of the repo.

FYI the `configure.in` and `*.m4` files work just fine for most cases but there are some issues that I have encountered i.e. compiling with Intel or linking to different PETSc libs. That possibly indicates the need update the configure scripts, but that is more challenging than it sounds since (usually) you have to maintain backwards compatibility.

adigitoleo:
I see, thanks for the info. This PR was clearly premature and I'll close it at least until I have understood more about autoconf etc. I don't trust myself to do an overhaul of the configure scripts just yet, and I can understand that it would be a delicate task.

PR #307, opened 2021-03-22
Moves the VTK dump after a solve.

### LINKED ISSUE #304: VTK state output out of sync with mesh adaptivity
I was wondering if anyone knows what is the benefit of outputting our VTK files in the next timestep iteration (i.e. at the top of the timestep loop) instead of outputting at the very end of a timestep right before the last exit condition (Fluids.F90:792).
Writing our VTK files in that order when adapting the mesh results into the solution from the current_time solve being output onto the newly adapted mesh on which we have not yet performed a solve.
Moreover, this setup for adaptivity, assuming that we output at each timestep, results into the last and second to last VTK outputs being on the same mesh which is not technically right.

In addition, the first VTK dump I believe should be moved to be just before `adapt_state_first_timestep` so as we ouput the input mesh on the zeroth timestep -> we then adapt it -> solve on it -> output timestep 1 mesh-> adapt mesh for next timestep -> top of the loop ->...

This can be seen for example in test: `adaptivity_preprocessing` where the last 2 timesteps have the same mesh even though we have gone and adapted (I have also turned on adapt at first timestep).

image

With the suggested changes (i have turned on adapt at first timestep for this) our VTK dumps would look like this

image

gnikit:
Fixes #304

This means that the solution is now displayed on the mesh on which it
occured and not on the interpolated mesh after a spatial adapt.

A `simulation_completed(current_time, timestep+1)` had to be used
for the cases where the option of `final_timestep` has been given to
prevent double output in the last timestep.

This still leaves the `adapt_at_first_timestep` since by the time the
first data dump occurs the mesh has already adapted, hence the 0th
and 1st timestep dumps look the same mesh-wise.

## TODO:

- Reorder the `adapt_at_first_timestep` to be after a data dump. That means that we need to:
 1. Obtain the timestep
   2. Move the check for VTK output
   3. Allocate and insert auxiliary fields
   4. Initialise the diagnostic fields
- I also think I need to delete some comments I added for the last time step data dump.

### DIFF
--- a/main/Fluids.F90
+++ b/main/Fluids.F90
@@ -421,26 +421,6 @@ contains
 
        call tic(TICTOC_ID_TIMESTEP)
 
-       if( &
-            ! Do not dump at the start of the simulation (this is handled by write_state call earlier)
-            & current_time > simulation_start_time &
-            ! Do not dump at the end of the simulation (this is handled by later write_state call)
-            & .and. current_time < finish_time &
-            ! Test write_state conditions
-            & .and. do_write_state(current_time, timestep) &
-            & ) then
-
-          !CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
-          ! Regular during run state dump.
-          !CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
-
-          ! Intermediate dumps
-          if(do_checkpoint_simulation(dump_no)) then
-             call checkpoint_simulation(state, cp_no = dump_no)
-          end if
-          call write_state(dump_no, state)
-       end if
-
        ewrite(2,*)'steady_state_tolerance,nonlinear_iterations:',steady_state_tolerance,nonlinear_iterations
 
        call copy_to_stored_values(state,"Old")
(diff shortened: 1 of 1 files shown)


main/Fluids.F90
```diff
@@ -839,6 +839,8 @@ SUBROUTINE FLUIDS()
     ! *** END OF TIMESTEP LOOP ***
     ! ****************************
 
+    ! Do not perform a dump if the timeloop exited because of the final_timestep
+    ! condition. That would result into the final timestep being output twice.
```
gnikit, line 843: delete

stephankramer:
> This means that the solution is now displayed on the mesh on which it
occured and not on the interpolated mesh after a spatial adapt.

I'm afraid this is a feature rather than a bug. I can absolutely see the arguments of doing it either way: it makes sense to want to see the solution on the mesh that it was solved on, it also makes sense to want to look at the solution immediately after an adapt: many applications have complex interpolation/projection procedures where the result isn't obvious and needs inspecting. Also the solution after interpolation is the information that moves forward into the next timestep, so if you dump every timestep you get to see the state immediately before a failed timestep. Bear in mind that in the applications of fluidity from this main branch typically don't adapt every timestep but typically say every 20 timesteps - and I get the impression it maybe isn't in your case?

Of course we could introduce an option that allows you to do it either way, but I'm not sure the default should change. In particular the default for when the first dump happens with adapt-at-first-timestep should be after the adapt: in a typical setup the input mesh is very coarse and seeing the fields which are all prescribed on that mesh is completely useless; typically the mesh is just coarse enough that adaptivity can still tell where to put more resolution (in a first iteration), and only when the mesh has fully developed is it possible to see the various features of the initial condition that form the starting point of the numerical simulation.

I'm also generally very wary about pulling apart anything to do with the ordering in which things are done in a timestep. This is a prime example of "brittle code" with fluidity having had too many and varied applications all implementing various whistles and bells, and diagnostics with complex interdependencies, that are also impossible to test in all combinations. So if you want to introduce an option to do something different that's absolutely fine, as long as it is obvious that it doesn't affect existing functionality.

gnikit:
> I'm afraid this is a feature rather than a bug. 

I see, this change was purely because of my work in radiation and what I perceive as optimal/logical. So I understand if there are cases where seeing the solution on the mesh on which the solve was performed might not be ideal.

> In particular the default for when the first dump happens with adapt-at-first-timestep should be after the adapt

I am actually very happy for you to say this because I won't have to mess with the adapt-at-first-timestep output. When I first started this issue, I tried to get it to work and for the life of me I couldn't, it was a true nightmare trying to initialise all the dependencies before the 0th output. So I am happy to drop that bit.

> So if you want to introduce an option to do something different that's absolutely fine, as long as it is obvious that it doesn't affect existing functionality.

That is completely understandable. I think it it only makes sense for me to add this only if you and the other devs think that this is a feature worth having in Fluidity, and it will be useful to someone down the road. Fluidity already has a massive code base and I wouldn't want to add to that features that make future maintenance harder. 

If you and/or other devs think that it makes little difference whether or not the solution is output before or after an adapt, especially since a data dump is normally performed every few timesteps, feel free to close this PR.

tmbgreaves:
Flagging that there is a proposed renaming of master branch to main on 26th May, which if this PR is still open at that point will require retargetting to main branch.

PR #311, opened 2021-04-22
Updates for Focal (Ubuntu 20.04) including gcc10 and updated CI

tmbgreaves:
This PR addresses:

- Fixes for building on Ubuntu Focal and Ubuntu Groovy
- CI migration to switch on testing on Focal and Groovy
- Test fixes on Bionic as a result of CI migration
- Test fixes for passing on Focal and Groovy

The substantial part of the PR is ready for review, with some (hopefully) minor test tweaks outstanding.

I'd welcome any thoughts and reviews on the CI migration from any/all developers - there is an ongoing push to try and move as many of the Fluidity codes as possible to Actions from Jenkins due to ageing Jenkins hardware and no clear way forward for the service. My goal is to try and get solid alternatives to Jenkins for all Fluidity codebases in place and switched to over the next month or two.


angus-g (commented):
This all looks pretty good to me! The Actions workflow is slick and seems to pick up failing tests just fine. I guess there would still be some tweaks needed to actually get the tests passing: things like the `particle_entrainment` test being (seemingly?) wildly off... It doesn't look like there were any substantial code changes, but I'm not really qualified to speculate beyond that!

tmbgreaves:
Thanks, @angus-g ! Feedback from your point of view as a developer on the Actions side is particularly appreciated. Rhod has the failing tests on his list to look at when time allows, hopefully over the next couple of weeks. I think the failures also exist in master, just not being tested, but it would be good to merge at a state where master tested with a full suite of passes.

examples/particle_rayleigh_taylor_mu10/particle_rayleigh_taylor_mu10.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <testproblem>
   <name>Particle-rayleigh-taylor-mu10</name>
-  <owner userid="cmath"/>
+  <owner userid="drhodrid"/>
   <tags>flml</tags>
-  <problem_definition length="medium" nprocs="1">
+  <problem_definition length="long" nprocs="1">
```
stephankramer, line 6: If it's changing to "long" it should move out into the longtests repo...

stephankramer (commented):
This is fantastic!

A few comments:
* if I look at https://github.com/FluidityProject/fluidity/pull/311/checks?check_run_id=2410218497 I have no way of telling whether these failed tests are in Bionic short or Bionic medium. Just knowing they failed isn't very informative in general. You want to go back to the log and search for the output of that test. So then I need to go to a local checkout to see what length the test is, which is complicated further here by the fact that the test actually changed length. Is what is displayed there based on the junit output? In which case we could add a short or medium tag to the junit output.

* the live-patch of h5hut during configure, isn't ideal as presumably it leaves you with modified source files which causes havoc with changing branches, rebasing, accidentally committing these changes etc. @angus-g : is this going to be fixed in h5hut itself? - should we just require hdf5 to be at least 1.12 ?

* quite a number of tests have gone from medium to short - which makes sense, but having "make test" take 2 hours isn't ideal either. We need a suite (maybe a new subcategory?) that can be run relatively quickly locally. Alternatively (or both) we should maybe consider a critical evaluation of which tests are still relevant and do a bit of a cull. Unfortunately that takes a considerable amount of time...

tmbgreaves:
Thank you @stephankramer ! Addressing your questions:

- Do we care whether tests are in the short or medium test suite from the point of view of failures? My thinking there was that they're all tests in tests/ and if anything fails we're likely to debug by pulling the test container and re-running locally, at which point the test length isn't particularly relevant and it's more useful to just have a combined 'this is what failed on operating system X' report. I'm open to being persuaded otherwise :-)

- The h5hut one is a horrible hack, agreed. I'm very open to a better fix there.

- Completely agreed on wanting a third category of tests. I'd found that even the ~30 minute short test suite was quite long for functionality checking and had wondered about having a 'functionality' tag with just a basic set which covered the critical things that go wrong with installs that are broken in the standard ways, and might indeed be the default target for 'make test' for basic acceptance. My view on the suites is that 'short' doesn't necessarily mean that the suite runs in short time, but that it contains tests that run in a short time, likewise medium.

examples/particle_rayleigh_taylor_mu10/particle_rayleigh_taylor_mu10.xml
tmbgreaves: Note that it's an example, not a test - so should probably stay where it is?

stephankramer:
So my workflow is to always go back to the log at the CI interface first. Quite often it's immediately obvious what's wrong and I can push a quick fix. Of course ideally I'd always rerun the test locally to confirm it fixes it, but some tests are a bit long, require some cores that I don't have on my laptop...and then there's the aspect of test results being different locally and in whichever build environment it's failing. You're right that we always have the docker container to fall back on for reproducability, but that tends to be my last resort really: it takes a fair while to download and development inside of it isn't ideal. Finally, an important aspect of looking at the logs online is that it allows you to compare the output between different runs. Sometimes the failure only gets noticed after a significant amount of development has passed. It can also help with intermittent failures...

So my ideal CI interface would have the entire test output right next to the failure message, but I'm also happy to continue my old routine of typing the test name in the search box of the relevant log, but then it does help to know which log I should be looking at. I'll have a look to see if I can fiddle with what we write into the junit output....

cianwilson:
This looks really great @tmbgreaves !  Really appreciated being able to look through the github workflows setup.

One quick question, if we're pushing at least 3 tags to dockerhub per pull request (more if the pr results in extra commits), is there some limit on how much we can push?  Are these tags auto-deleted eventually?

jrper:
As a point of information for @stephankramer, what's appearing is indeed the JUnit test name (or rather the test case name), which is fairly easily mangleable. I _think_ the information you want is stored as the test suite name, but it's too long since I looked at it to be sure. Depending on what Actions shows, we might be able to stick the log as the error message so it's more immediately visible, although I vaguely remember the testharness parallelism introducing a complication to tha5 approach.

tmbgreaves:
> So my ideal CI interface would have the entire test output right next to the failure message, but I'm also happy to continue my old routine of typing the test name in the search box of the relevant log, but then it does help to know which log I should be looking at. 

Very useful feedback, thank you @stephankramer . I think I can see how to easily split apart the short and medium junit reporting - will have a go at that now. 

The JUnit reporting action is one I picked up from the marketplace, and might well be able to be improved significantly for what we're doing.

tmbgreaves:
Ok - 3f57d07e4 changes the behavior of Actions back to the earlier option of failing the test run if the test run itself and/or the tests experience failure. This should make it easier to pinpoint where the relevant log files are. Once that run has gone through, @stephankramer , might you be able to take a look and give some feedback on whether it's an improvement?

The actual error is still in the logs, which I guess is no worse than the Jenkins situation, but very much agree that having it come out in the JUnit report would be ideal.

tmbgreaves:
> One quick question, if we're pushing at least 3 tags to dockerhub per pull request (more if the pr results in extra commits), is there some limit on how much we can push? Are these tags auto-deleted eventually?

I don't think there is a push limit - at least, not one I've come across. There is a pull limit for non-authenticated systems, of 100 pulls per six hours, which I think we're nowhere near. In terms of having lots of tags, I'm pushing to fluidity/actions which is apart from the normal fluidity/baseimages which the build containers go to, so hopefully the profusion of tags won't be a problem. My guess is that apart from Actions, the only pulls will be people doing a debugging run and grabbing a tag from the Actions logs to pull, rather than wading through the forest of tags in fluidity/actions .

I think dockerhub auto-deletes images which haven't been updates/accessed in six months (?) which should act as a backstop cleanup for us.

tmbgreaves:
Possibly also worth adding for background - I had initially planned to just have the tests run in containers built per-test-run rather than building and cacheing on dockerhub, but ran into the annoying problem that github don't allow any dynamic content in that section, as per:

https://github.community/t/how-to-use-env-with-container-image/17252/4

So rather than use a github method, 'for security reasons' I end up using my own hacked-up method via dockerhub....

tmbgreaves:
@stephankramer a first run with the updated reporting:

https://github.com/FluidityProject/fluidity/actions/runs/778948034

tmbgreaves:
@stephankramer a run to review with the updated reporting:

https://github.com/FluidityProject/fluidity/actions/runs/779218872

Patol75:
Regarding why failed tests only display "Failure" as an annotation, I think this comes from [testharness](https://github.com/FluidityProject/fluidity/blob/6df2b96005b7a1b45fce5e502726bd670aff3707/tools/testharness.py#L341). The message argument of [add_failure_info](https://github.com/kyrus/python-junit-xml/blob/4bd08a272f059998cedf9b7779f944d49eba13a6/junit_xml/__init__.py#L423) seems to be the one picked up and displayed in Action. For example, if you compare with [unittestharness](https://github.com/FluidityProject/fluidity/blob/6df2b96005b7a1b45fce5e502726bd670aff3707/tools/unittestharness.py#L134), message there contains an error message rather than just "Failure".

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: ${{ matrix.name }} Testing
+        run: |
+         chmod 777 .
+         docker pull fluidity/actions:${{ matrix.release }}-${{ github.sha }}
+         docker run -v $PWD:/host fluidity/actions:${{ matrix.release }}-${{ github.sha }} /bin/bash -c "${{ matrix.command }} && cp -v tests/${{ matrix.output }} /host/${{ matrix.release }}-${{ matrix.output}}"
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ${{ matrix.release }}-${{ matrix.output }}
```
Patol75, line 120: You could potentially add `name: tests_xml_outputs` to make it more obvious what this "artifact" archive contains.

tmbgreaves: Good catch! I'd missed that that was lacking a name - fixed in ef3a9ba .

tmbgreaves:
Noting @drhodrid 's proposed fixes in PR #313 as relevant to this PR.

stephankramer:
> @stephankramer a run to review with the updated reporting:
> 
> https://github.com/FluidityProject/fluidity/actions/runs/779218872

That's great @tmbgreaves , many thanks - all happy now

tmbgreaves:
Attempting to open up the possibility of adding some longer tests in the standard run, 45327d8 pushes some long tests which should complete on github-hosted runners within the time that the mediumtest suite takes, so not slowing down the overall run assuming github gives enough resources to run everything at the same time. The benefit is having some longer tests run against PRs to master, the cost is complicating the set of outputs from the run.

@drhodrid - I'll also post this to your PR, but if you can add your tests which have gone over the mediumtesting length to the longtests repo and then add them in to the longtesting matrix set up in 45327d8 , I propose that's how we keep testing going but don't overall lengthen the runtime.

Patol75:
As we are dealing with tests in this PR, is it normal that `tests/python_shape_dshape_test` and `tests/python_shape_dshape_test_2d` only have a `Makefile` in their directory? It looks like the Makefile is trying to access a`src` directory that does not exist.

stephankramer:
Yes looks like these tests were removed in f5604ccb47c284aa470ab91e69d8a4a8af67a626 but their Makefiles were updated on a parallel branch in 06a11efe035ec821dc7aa23f12d4855b83adebd4 so got accidentaly added back in upon merge. Removed now...

tmbgreaves:
@stephankramer - I'd very much value your thoughts on my suggested addition of some longtests to the standard master/PR runs, and how it's being done/presented.

tmbgreaves:
I'm increasingly unconvinced that I'm going about implementing longtesting the right way here. I think a better method by far may be to have the testing happen separately and only on pushes to master, as this is rapidly turning into a resource black hole which is eating up our allocation of runners each time a PR is updated. On the one hand we could take the view that running tests for PRs going into master just takes a long time and in the process makes sure that we check changes to master as robustly as possible, on the other hand we could take the view that mediumtesting catches pretty much everything we care about, and the longtests are an 'after the push' check to make sure nothing slipped through the mediumtest net that needs hotfixing.

Opinions on a postcard, please....

tmbgreaves:
Now in what is hopefully the tail end of the CI overhaul and attempting to get longtests / examples being properly tested, some documentation of the remaining failures:

Examples:

* driven_cavity : looks like one (hopefully tolerance) failure
* hokkaido-nansei-oki_tsunami :
* backward_facing_step_2d : 

Longtests:

* backward_facing_step_2d :
* coarse-corner :
* explicit-hyperc-shear :
* explicit-hyperc-shear-adapt :
* explicit-hyperc-superman :
* gls-StationPapa :
* lagrangian_detectors_3d_2e5 :
* lock_exchange :
* lock_exchange_2d_Lagrangian_paths :
* lock_exchange_3d_parallel :
* medp1dgp2 :
* particle_entrainment_of_dense_layer :
* saltfinger2d_adaptive :
* spherical_* : wierdly, passing for me locally but failing on github runners; trying to figure out why...
* square-convection-1e6 :
* Stokes_mms_p1dg_p2 :
* Stokes_subduction_zone_vanKeken2008_OneA :
* Stokes_subduction_zone_vanKeken2008_TwoB :
* wetting_and_drying_balzano1_dg_parallel"

VLong tests/example still to be set up:

* tides_in_the_Mediterranean_Sea : 
* restratification_after_oodc : 
* mms_ns_p1dgp2_steady_periodic_parallel : 
* mms_ns_p1dgp2_steady_periodic : 
* mphase_tephra_settling_salinity : 
* mphase_mms_p2p1_compressible : 
* sphere-3D-drag-Re1000 : 
* backward_facing_step_2d_longtest : 
* backward_facing_step_3d_longtest : 
* mphase_tephra_settling_salinity_adaptive : 
* flredecomp_long : 
* mphase_rans_p1dgp2_cv_bubble_column_2D : 
* mphase_mms_p0p1cv_periodic : 
* flow_past_a_square : 
* mms_ns_p2lp1_supg_steady_compressible : 
* lituya_bay_parallel_debug :

tmbgreaves:
The gcc-fixes-plus-longtests branch is currently processing the remaining prior-to-this-overhaul-untested failures, and I'll aim to trial a run of the vlong tests (probably via the old Jenkins builders) there too to complete the testing suite.

tmbgreaves:
Testing of vlong tests initiated at : https://github.com/FluidityProject/longtests/actions

This will almost certainly be buggy at first. It's running on Jenkins hardware.

tmbgreaves:
@drhodrid - Almost there with your tests, thank you! Last one, hopefully a tolerance issue, in particle_prescribed_vortex_flow running on Focal I'm seeing:

Max cons error 0.010923752666083475

against a test of 'less than 0.01'. Can this be relaxed to pass?

tmbgreaves:
@stephankramer @jrper - wondering if you have any thoughts on what might be going on with the one wierd remaining short test failure of square-convection-parallel-trivial ? 

What I'm seeing is that it fairly consistently segfaults when running in the short test suite on github runners but so far has never segfaulted in the same container run on my local system. I've tried setting it up on a separate repo and have had it both segfault and run successfully on github runners, sometimes on sequential runs in the same container. I've not yet managed to catch it segfaulting to get a backtrace out of it.

Does this touch anywhere unusual in the codebase that might be a culprit, or do anything unusual?

I'm out of ideas...

tmbgreaves:
@matt-piggott - I'm seeing what I hope is a marginal test failure in the driven_cavity example on Ubuntu 20.04, where a test of convergence > 1.85 fails on:

driven_cavity: Running Quantity4ConvergesToSecondOrder:
[[1.72761663]
[2.23112378]
[2.61936127]]
driven_cavity: failure.

Can this be relaxed slightly?

tmbgreaves:
@angus-g - I think I have a question for you: it looks like some of the examples / longtests which parse detectors are failing as a result of detector files now outputting as h5part, but the examples/tests are still configured to read old-style detectors output via stat_parse from the fluidity_tools module. On a first pass, tests which I suspect are failing from this are:

* wetting_and_drying_balzano1_dg_parallel 
* hokkaido-nansei-oki_tsunami
* lock_exchange_2d_Lagrangian_paths 
* square-convection-1e6

Does this sound like it could be a reasonable diagnosis and if so how straightforward would a fix be?

tmbgreaves:
@jhill1 - I'm sweeping through the longtests in an attempt to get them running in a sensibly controlled environment (Ubuntu 20.04 container rather than ICHPC...) and have what I hope is a simple failure to fix in medp1dgp2, namely a report from flredecomp of:

ERROR (Calendar.cpp): The _from_unit specification contains a syntax error.

I'm guessing this is coming from changes to utScan in recent UDUnits2 failing to parse <time_units date="seconds since 1987-01-05 00:00.0"> from the flml.

This should reproduce in a recent Fluidity container, ie:

```
docker run -it fluidity/actions:focal-5ec2aa3ac9fa468daccff3a84e6272d87244dc52
git clone https://github.com/fluidityproject/longtests
bin/testharness -f medp1dgp2.xml
```

I'm hoping you might be more clued up on utScan input syntax than me and be in a position to put a quick fix in here .... ? :-)

drhodrid:
Hi Tim,

Yes - I think you can safely relax this one.

R

On 5 May 2021, at 8:22 am, Tim Greaves ***@***.******@***.***>> wrote:



@drhodrid<https://github.com/drhodrid> - Almost there with your tests, thank you! Last one, hopefully a tolerance issue, in particle_prescribed_vortex_flow running on Focal I'm seeing:

Max cons error 0.010923752666083475

against a test of 'less than 0.01'. Can this be relaxed to pass?

stephankramer:
@tmbgreaves : re. running the longtest as well. Do I understand correctly all these are now separate jobs that in principle could run in paralllel? If so, yes sounds great - and it looks good from the interface. I guess it depends a bit whether this is going to affect the overall completion time before reaching the green light...

I'll take a look at square-convection-parallel-trivial to see if can reproduce...

tmbgreaves:
@stephankramer - yes, that's exactly what I'm aiming for in terms of them running in parallel. At the moment I'm seeing a *bit* of slowdown on the run as a whole - perhaps adding ~2.5h to the already ~5h run, which may extend a little as the last few longtests/examples are fixed and added back in. The limit here is the 20 github runners we get at a time. My feeling is that it's worth it for the hopefully significantly better coverage we get by regularly running most of the longtests against PRs.

I think the time-to-green-light is probably not too significant on the timescale of getting code reviews etc.

There is also the 'vlong' suite of tests which I didn't manage to fit onto github runners, likely going to be on the order of 20 from examples and longtests, and which I'll aim to run via the longtests repo on self-hosted hardware in a similar weekly schedule to the way we used to run all longtests, but now in a much more controlled environment.

I'd suggest that if we can get to the point that square-convection-parallel-trivial is passing we move in the direction of final reviews and (hopefully) merging this PR with whatever tests are passing, given that should then be overall green, and we can then drip-feed in the remaining longtests as and when they're fixed.

stephankramer:
All sounds good re. longtests - we're not in a situation where we have many active PRs in parallel - so that still sounds manageable

Re. the square-convection-parallel-trivial test case: I do have a backtrace now. It's segfaulting in the packing of detectors in parallel, line 455 of Detector_Tools.F90. Unfortunately it seems a bit heisenbuggy - if I rebuild with debugging it doesn't hit the segfault - and some of the variables I'd like to inspect are `<optimized-out>`. @drhodrid - in case it's relevant to failures you're looking at...

Slightly off-topic @tmbgreaves  - but I was enjoying the flawless reproducability by running the relevant docker image. It seems the docker image is very bare bone (which is good!) - but it would be nice if the fluidity user could have `sudo` access so I can install stuff in a debugging session. At the moment it doesn't seem to have any editors, not even vi!

tmbgreaves:
Thank you very much for the debugging on square-convection-parallel-trivial, @stephankramer . I don't think there's much I can do, but if there is, do shout.

I was waiting for someone to request more useful features in the container, @stephankramer , and being lazy implementing until someone did ;-) I'll add in sudo now - that's a very good idea.

tmbgreaves:
Sudo available in containers from 48e1d2abb (ie, next build of this PR).

tmbgreaves:
It's spurious pending fixes to square-convection-parallel-trivial but Actions has now completed a fully-passing run which took just under 7.5h, confirming the <3h additional time for the longtest suite running thus far.

stephankramer:
Yes, happy if you want to go ahead. Progress on square-convection-parallel-trivial in separate issue #317 I think it's a real issue, but likely already in master...

PR #320, opened 2021-05-18
Allocate empty attribute arrays for detectors

### LINKED ISSUE #317: Segfault in square-convection-parallel-trivial
This test segfaults in the last adapt when trying to pack data for detector migration in zoltan
Backtrace:
```
#0  0x000055b2ee6ed510 in detector_tools::pack_detector (detector=0x55b2f0ac7460, buff=..., ndims=<optimized out>, 
    nstages=<error reading variable: Cannot access memory at address 0x0>, attribute_size=...) at Detector_Tools.F90:452
#1  0x000055b2eed4a365 in zoltan_callbacks::zoltan_cb_pack_fields (data=..., num_gid_entries=<optimized out>, num_lid_entries=<optimized out>, num_ids=<optimized out>, 
    global_ids=..., local_ids=..., dest=..., sizes=..., idx=..., buf=..., ierr=0) at Zoltan_callbacks.F90:1101
```
which is
```
buff(ndims+det_params+1:ndims+det_params+attribute_size(1)) = detector%attributes
```
Looks like the lhs is fine - it's actually the `detector%attributes` that is unallocated (`print *, allocateddetector%attributes` print F). This is also flagged by valgrind first in line 56 of Zoltan_detectors.F90:
```
   ! loop through all registered detector lists
    call get_registered_detector_lists(detector_list_array)
    do det_list = 1, size(detector_list_array)

       ! search through all the local detectors in this list
       detector => detector_list_array(det_list)%ptr%first
       !Set up particle attribute parameters
       total_attributes=0
       if (associated(detector)) then
          if (size(detector%attributes)>=1) then
```
So the detector appears associated, but asking for the size of `detector%attributes` is flagged by valgrind as using uninitialised memory.

angus-g:
(Duplicated from #319)

When we allocate/unpack detectors from Zoltan, some code expects that their attribute arrays (used for particles) are at least allocated, even if length 0. Ensuring this is done avoids potential segfaults when these arrays are unconditionally accessed.

Thanks @stephankramer for the detailed analysis of this issue! I think this is probably the right place for this allocation? Of course, the other approach would be to figure out where attribute_size is being passed into unpack_detector as a non-optional on a detector. (As an aside, that subroutine looks like it could just take a non-optional attribute_size and then the logic wouldn't be exactly duplicated in both branches...)

Closes #317.

### DIFF
--- a/assemble/Zoltan_callbacks.F90
+++ b/assemble/Zoltan_callbacks.F90
@@ -1081,24 +1081,20 @@ contains
        ! packing the detectors in that element
        do j=1,zoltan_global_ndets_in_ele(i)
           !check attribute sizes
-          if (size(detector%attributes)>=1) then
-             attribute_size(1)=size(detector%attributes)
-             attribute_size(2)=size(detector%old_attributes)
-             attribute_size(3)=size(detector%old_fields)
-          else
-             attribute_size(1)=0
-             attribute_size(2)=0
-             attribute_size(3)=0
-          end if
+          attribute_size(1) = size(detector%attributes)
+          attribute_size(2) = size(detector%old_attributes)
+          attribute_size(3) = size(detector%old_fields)
+
           total_attributes = sum(attribute_size)
           !pack attribute sizes
           do k = 1,3
-             rbuf(rhead)=attribute_size(k)
-             rhead=rhead+1
+            rbuf(rhead)=attribute_size(k)
+            rhead=rhead+1
           end do
+
           !pack the detector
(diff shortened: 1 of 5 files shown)


stephankramer (dismissed):
Yes, looks right to me. Thanks!

stephankramer:
Oh except there are now new (I believe?) particle test failures...

angus-g:
Ah yes, I was going to ask whether those were existing or not...I might have to dig a bit deeper into what's going on!

tmbgreaves:
All green on the PR now, thanks @angus-g :-)

@stephankramer - might you be able to review again please and merge if you're happy?

stephankramer (approved):
All looking great - thanks!

Issue #315, opened 2021-05-04
Segfaults in flredecomp from longtests


tmbgreaves:
Longtests are showing reproduceable segfaults in flredecomp, so far exhibiting fairly quickly in:

* lagrangian_detectors_3d_2e5 
* Stokes_subduction_zone_vanKeken2008_TwoB

Example from lagrangian_detectors_3d_2e5 : 

```
[4b33659a299d:00363] *** Process received signal ***
[4b33659a299d:00363] Signal: Segmentation fault (11)
[4b33659a299d:00363] Signal code: Address not mapped (1)
[4b33659a299d:00363] Failing at address: 0x65
[4b33659a299d:00363] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7fb4089ae3c0]
[4b33659a299d:00363] [ 1] /lib/x86_64-linux-gnu/libc.so.6(+0x18e7a3)[0x7fb4075f77a3]
[4b33659a299d:00363] [ 2] /lib/x86_64-linux-gnu/libgfortran.so.5(+0x268295)[0x7fb408c3f295]
[4b33659a299d:00363] [ 3] /lib/x86_64-linux-gnu/libgfortran.so.5(+0x268574)[0x7fb408c3f574]
[4b33659a299d:00363] [ 4] flredecomp(__checkpoint_MOD_checkpoint_detectors+0x297c)[0x561d5660ca8c]
[4b33659a299d:00363] [ 5] flredecomp(__checkpoint_MOD_checkpoint_simulation+0x3cc6)[0x561d56610866]
[4b33659a299d:00363] [ 6] flredecomp(flredecomp+0x811)[0x561d566067e1]
[4b33659a299d:00363] [ 7] flredecomp(main+0x66f)[0x561d566049ef]
[4b33659a299d:00363] [ 8] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7fb4074900b3]
[4b33659a299d:00363] [ 9] flredecomp(_start+0x2e)[0x561d56605f0e]
[4b33659a299d:00363] *** End of error message ***
```

Also seen in Actions, eg: https://github.com/FluidityProject/fluidity/runs/2495251930?check_suite_focus=true

Reproduce with:

```
docker run -it fluidity/actions:focal-686bb4ef5cd8e079edfff77598b7e55d7d117547 bash -il
git clone https://github.com/fluidityproject/longtests
[...]

PR #321, opened 2021-05-19
Improved error messages from python

stephankramer:
Often when an error occurs in the python embedded in fluidity, the
error+backtrace would not show up despite use calling PyErr_Print and
only prints the useless FLAbort message "Dying!". This
is caused by some internal buffering in python of stdio and stderr
(internal to python so independent of whether we have redirected the
actual stdio/stderr of the process at the system level). This commit
switches all such buffering off. As suggested here:
https://stackoverflow.com/questions/29352485/python-print-not-working-when-embedded-into-mpi-program/49076389#49076389
under option 1) Note that it should only affect stdio/err and not all
I/O from python (as it claims on stackexchange), and since we
shouldn't normally be printing much from python this should hopefully
not affect performance.

Also fixed a case where the error would result in a segfault because we
try to decref the pyobject returned from PyRun_String before checking
whether an error has occured, in which case PyRun_String returns null.

This came up in #315

### DIFF
--- a/femtools/embed_python.c
+++ b/femtools/embed_python.c
@@ -65,7 +65,15 @@ int eval_user_func(char *function, int function_len, PyObject *pLocals, PyObject
 
   // Execute the user's code and clean up allocated string.
   pCode = PyRun_String(function_c, Py_file_input, pGlobals, pLocals);
+
   free(function_c);
+
+  // Check for errors in executing user code.
+  if (PyErr_Occurred()) {
+    PyErr_Print();
+    return 1;
+  }
+
   Py_DECREF(pCode);
 
   // Extract the function from the code.
@@ -75,11 +83,6 @@ int eval_user_func(char *function, int function_len, PyObject *pLocals, PyObject
       return 1;
   }
 
-  // Check for errors in executing user code.
-  if (PyErr_Occurred()) {
-    PyErr_Print();
-    return 1;
-  }
 
   return 0;
 }
--- a/femtools/python_statec.c
+++ b/femtools/python_statec.c
@@ -21,6 +21,10 @@ void python_init_(void){
 #if PY_MAJOR_VERSION >= 3
   PyImport_AppendInittab("spud_manager", &PyInit_spud_manager);
 #endif
+  // Force the stdout and stderr streams to be unbuffered
+  // This is to ensure backtrace+error actually gets printed
+  // through PyErr_Print before we abort.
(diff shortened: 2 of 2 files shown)


jhill1:
Nice. I don't see any major issues with this and well found on the fix!

angus-g (approved):
Nice fix on preventing the segfault, that was a big blocker for sure!

I think the `Py_UnbufferedStdioFlag` fix seems to be fine, although maybe it does different things depending on the Python version? Basically we should be getting _some_ sort of buffering now, and it can be either straight write-through or line buffering.

Prior to 3.7, the logic in the linked Stack Overflow thread holds, i.e.
```python
    if (isatty || Py_UnbufferedStdioFlag)
        line_buffering = Py_True;
    else
        line_buffering = Py_False;
```

This changes in 3.7 (https://github.com/python/cpython/pull/1667):
```python
    if (Py_UnbufferedStdioFlag)
        write_through = Py_True;
    else
        write_through = Py_False;
    if (isatty && !Py_UnbufferedStdioFlag)
        line_buffering = Py_True;
    else
        line_buffering = Py_False;
```

And back again in 3.9 (https://github.com/python/cpython/pull/17646):
```python
    if (!buffered_stdio)
        write_through = Py_True;
    else
        write_through = Py_False;
    if (buffered_stdio && (isatty || fd == fileno(stderr)))
        line_buffering = Py_True;
    else
        line_buffering = Py_False;
```

stephankramer:
Ah great, thanks for digging that up. Right, so if I understand correctly there's basically three different modes 1) block buffering (which is the normal default for all files) 2) line buffering 3) immediate write through. I think `Py_UnbufferedStdioFlag` was intended (in python 2) to give you write through, but according to the discussion that wasn't available in early python3. So that's why prior to 3.7 it just switched to line buffering instead (instead of block buffering). Then in 3.7 they realized the functionality was now available so you do get write through with that flag now. I don't think anything changes in 3.9 if you have that flag set, I presume it means unbuffered_studio will be false. It's just that in 3.9 without that flag it will always use line buffering (instead of block) for stderr - even if it's not a tty. The issue we were having is that in parallel stderr fails the `isatty` test even if we don't redirect ourselves (presumably because it's redirected by mpi before being dumped in the terminal). For our purpose I guess line buffering should be sufficient which is what we get by default in 3.9 - but for older versions there seems to be no simple way to achieve that.

PR #325, opened 2021-06-01
Examples: Delete hokkaido-nansei-oki_tsunami

tmbgreaves:
This example is not well tested within the current Fluidity framework, failing current tests, and not considered worth spending the time to update at present by the core developer team; see discussion in issue #315


tmbgreaves:
Tests not run on this PR as it's only removing an example, but opened as a PR to highlight the proposed example removal so if anyone has an opinion either on the removal or on fixing the tests there's a place to discuss.

stephankramer:
I'm all for it - unless somebody wants to step up (presumably someone who cares about tsunami modelling in fluidity), and convert it and make it a proper test: shouldn't be too much work tbh. If we do remove it, should we remove it from the manual as well?

jhill1:
Happy to remove. I'm not going to have time to look at this in a reasonable timefeame. If I do get around to looking, I can add it back in.

tmbgreaves:
Thanks all - I've removed the relevant section from the manual, and will let it run through checks to make sure I've not broken anything unintended in the process. Assuming all is good I'll merge tomorrow - if anyone wants to put in some time to fix this up, the two commits should be easy to revert.

PR #327, opened 2021-07-15
Fix "mg" preconditioner with petsc 3.14

stephankramer:
This is dealing with bug https://gitlab.com/petsc/petsc/-/issues/868
Which is fixed only for 3.15. Unfortunately no workarounds for 3.14 exists in fortran,
because even if we do provide the optional argument it still gets
recognized for NULL and hits the bug. So instead write our own wrapper
in C.

This also drop support for petsc version <3.8 Note that Bionic official
packages are 3.7 - but we have always supplied our own 3.8 package for
it. Groovy and Hirsute are now on 3.14 Dropping support for <3.8
simplifies/removes a lot of the legacy workarounds. We no longer support
linking with petsc without fortran modules (we never did for versions >=3.8).


angus-g (changes requested):
Nice, it's so much cleaner to remove all those PETSc version checks scattered around. I think the `--disable-petsc-fortran-modules` argument can be removed:
https://github.com/FluidityProject/fluidity/blob/c1fc8c4fd569deb234bc83f7725988cc3a0c356f/configure.in#L831-L833? And `configure` will also need re-generation, because the PETSc version check is still looking for >=3.4.

angus-g (dismissed):
LGTM now!

angus-g (approved):
Still LGTM, ready to merge?

stephankramer:
Thanks @angus-g

PR #333, opened 2021-08-04
Allow and test coupled_cv with moving meshes

cianwilson:
Previously the coupled_cv assembly didn't take account of the grid velocity.  This fixes this in a manner that is equivalent to the normal cv assembler.

It also adds two tests.  One which tests that this is working with conservative advection and one that tests nonconservative advection with a divergence free (in the cv dual space) velocity field.


drhodrid (dismissed):
This works very well for me and the code changes and tests look great.

cianwilson:
Thanks @drhodrid !  I just rebased the branch onto the current main in preparation for merging.  Will let it rerun tests first.

stephankramer (approved):
Looks perfect. Didn't check the maths in detail but looks plausible and well tested

PR #334, opened 2021-08-09
Parallel non extruded ocean boundaries

stephankramer:
This PR includes the following:
* allow vertical extrapolation with non-extruded meshes in parallel. The implementation is not very efficient as it requires interpolation from the surface mesh to the volume mesh, which are decomposed independently in parallel. The assumption is that the surface mesh is relatively small, so rather than having all processes send the coordinates of all their volume nodes to all owners of the surface mesh, the surface mesh is duplicated such that each process can see the entire global surface mesh and only values on the surface mesh are distributed (broadcast to all other processes). This functionality is required to run with (non-extruded) free surface mesh movement in parallel

Other fixes related to free surface mesh movement in this PR (originally in #130):
* enable adaptivity+free surface mesh movement. This is done by moving the mesh back to `OriginalCoordinates` before the adapt so that the domain with the (assumed flat) initial free surface is adapted. After the adapt, the mesh is moved back according to the interpolated free surface field
* allow a density that is not one in the viscous free surface (with LinearMomentum)

Fixes to particles with a moving free surface: 
* they need relocating after each mesh movement
* in combination with adaptivity, we need to ensure that the particles stay within their element when we move the mesh back to its original geometry before the adapt (so zoltan can correctly deal with them), and move back with the mesh when the mesh movement is re-applied after the adapt.


cianwilson:
This all looks good!

My only question would be whether we should add the `beaker_blob[_adapt]` tests with the owner currently set as `bsnelling`, this appears to be a layover from the 2016 PR #130 .

Also, the `beaker_blob` test is surprisingly slow.  Thought it was just 1D but turns out it's extruded and ends up being quite high resolution.  Is that necessary or could the vertical resolution (at least) be lowered a little?  The `_adapt` version starts out with the same mesh but is much better post adapt.

stephankramer:
Thanks, @cianwilson - much appreciated. I've changed ownership of these tests. On  my laptop beaker_blob runs in just 46s. I'd be interested to know if it's much slower for you cause that could indicate solver issues...

cianwilson:
Not horrendously slow admittedly.  1:30 for me on `beaker_blob`.  But when I look at it, it's quite high resolution.  It fails some tests at lower resolutions but as far as I can tell it appears to be a regression test (and sometimes testing things that aren't even guaranteed to be true, like conservation of the volume fraction) so I was wondering if the resolution could be dropped and the regression numbers adjusted.

cianwilson (dismissed):

cianwilson (approved):

PR #336, opened 2021-09-12
Minor bugs spotted in Focal

gnikit:
While doing some work I encountered one more short-circuit evaluation in Sediments along with some compiler failures
when enabling for using recursion in routines not marked as such (you can replicate the failures by turning -fcheck=recursion` on during compilation).

I also fixed some of the `clean` targets in our test Makefiles because they were polluting the git tree after running tests.

## Summary of changes

1. Fixes short circuit evaluations in sediments
2. Adds missing recursion keyword for subroutines that are used in recursion
3. Edits some make files in tests to completely clean the test output
4. Made `.gitignore` slightly more aggressive to specifically ignore all the configure and compile generated files and nothing more

I believe this is now done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,12 @@
 ## Linux related
 *~
 
+## VSCode
+.vscode
+
+## Settings for the Fortran Language Server
+.fortls
+
 ## Emacs files
 # -*- mode: gitignore; -*-
 \#*\#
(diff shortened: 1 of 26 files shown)


stephankramer (approved):
All good, thanks!

Out of interest: did you run the entire test suite with `-fcheck-recursive` ? Maybe we should add it to our debug flags?

gnikit:
I did and I haven't seen any other issues, actually I config with `-fcheck=all,no-array-temps`, which takes care of a few more things like pointers, mem allocation, out of bounds checks, etc. I agree it might be useful when debugging.

gnikit:
The only other thing that I would ideally add in this PR is to make all the regression tests use a relative path to fluidity/flredecomp instead of just calling fluidity i.e. `../../bin/fluidity` in place of `fluidity` because testing between multiple versions of Fluidity becomes harder than it should be (IMO).

Patol75:
On this last comment, that could be easily done by adding `replace` methods [here](https://github.com/FluidityProject/fluidity/blob/068bdc5da5bd79aba83a95e21c5806ee41e61a08/tools/testharness.py#L71-L72).

gnikit:
@Patol75 you are right, I think it might be worth doing this in a separate PR to keep things tidy. 
Might also be worth updating the .flml and .xml files with spud-update-options, replacing the XML comments with <comment></comment> blocks, etc.

gnikit:
@stephankramer would you mind having one last look (I forgot to push a couple commits last time) so the review was auto-rejected.

stephankramer (approved):

PR #337, opened 2021-09-15
Adding use statement to compile this

jhill1:
Really minor change so cgal will compile. Not tested running the code.


Patol75:
@stephankramer Can this one be merged considering CI returned green?

stephankramer:
Yeah I think so. I guess this means we're not testing with CGAL at the moment, so no idea whether this has helped - but (after rerunning make makefiles) it shouldn't have an effect on the normal no-CGAL case.

stephankramer (approved):

PR #348, opened 2022-01-17
Account for distutils being deprecated and python3.10 becoming default on Ubuntu 22.04

### LINKED ISSUE #346: Replace distutils by setuptools
Just a heads-up that Python 3.10 featured [PEP 632](https://www.python.org/dev/peps/pep-0632/), which deprecates the distutils package from the standard library. In this newer version, the interpreter returns:
```
>>> from distutils.core import setup
<stdin>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```

Patol75:
Closes #346

I am opening this PR as a draft; it will be needed in the future given changes in the Python distribution. Please feel free to contribute, as it is likely I will succeed in figuring everything out.

I have tried to identify instances where `distutils` is used and replace them with equivalent logic from either `setuptools` or `sysconfig`.

Something of potential interest is `setuptools` indicating that the `setup.py` approach is deprecated. Instead, `pyproject.toml` and `setup.cfg` files should be used, in combination with the `build` and `pip` packages. I have tried to build the `fluidity` package this way.

---

Judging from Actions, Focal is successful, but Bionic does not find a valid version of Python.


Patol75:
I think this is starting to head in the right direction. I am giving it a CI test to see if the new `python_build` behaves.

@stephankramer
- Could you let me know what the `aclocal.m4` files are and what they do? I have modified them to mirror the changes in the `configure` files, but it is not clear to me what this actually does.
- To account for the deprecated `setup.py` approach and the new recommended `build` + `pip` strategy, I have removed `default` and `fltools` from the `install` target in `Makefile.in`. The main reason is that `pip` should not be run as root (I still do it in the `clean` target, but I guess this is more acceptable). Considering `default` and `fltools` do not require root privileges, but `install` does, I am not sure if it made sense to have them in. Additionally, an improvement from the new `pip` logic is that `PYTHONPATH` does not need to be set anymore to access fluidity's python modules.
- Some C++ files in `tools` were apparently missing the `unistd.h` header inclusion. I am not quite sure how `chdir` was executed before.

gnikit:
@Patol75 you might want to have a look at https://github.com/nektos/act . I use it for local Actions testing. Maybe you can use it with only a couple failing tests, on a single docker container and get some info on the errors.

Also, I am not sure if this is relevant but `HAVE_NUMPY` if true, is never actually `#define`d in `confdefs.h` like the rest of our preprocessor macros, so that might lead to some cryptic failures.

Patol75:
Thanks for the suggestion. I tried, but I did not manage to get a successful build, unfortunately. I thought `fluidity` being a sudoer was the reason that Python could not find installed libraries, but apparently not. It is annoying because I cannot reproduce locally...

Patol75:
I am closing this PR as I will likely open a new one based on the [deprecate_distutils_2](https://github.com/FluidityProject/fluidity/tree/deprecate_distutils_2) branch.

PR #352, opened 2022-01-28
Remove Groovy from CI

stephankramer:
Groovy has reached end of life.


gnikit:
@stephankramer I think you need to also remove the branch protection rule for Groovy for it to stop being required.
>Settings -> branches -> Branch protection rules

Patol75 (approved):

PR #353, opened 2022-03-13
BUG: Fix paths for generated headers for out-of-source builds

adigitoleo:
By default, `configure_file` looks for the input files in
`CMAKE_CURRENT_SOURCE_DIR`, but the generated headers are placed in
`CMAKE_BINARY_DIR`.

In the case of out-of-source builds the specified /include dirs also don't exist.
This causes cmake to treat /include as a filename instead of a directory.
The commit changes `configure_file` directives to always specify a _file_ path.

To reproduce the bug (run in source tree root):

```
mkdir build
cd build
cmake ..
```


adigitoleo:
Will send to Thomas as patch instead.

PR #354, opened 2022-05-25
ENH: Update VTK Ghost API

Patol75:
This PR is a port of changes originally written for this [commit](https://github.com/FluidityProject/fluidity/commit/75d59c6c06c1a0dadc485f525f9f9024c133c922); hopefully, I did not break too many things in the process.

The purpose of this PR is to reflect already [old changes in the VTK API](https://www.kitware.com/ghost-and-blanking-visibility-changes/) regarding ghosts. Currently, loading a Fluidity PVTU on recent versions of ParaView or using VTK bindings for Python will generate as many error messages as CPU cores were used, which is not a good sign. Furthermore, pvtu2vtu is broken when attempting to generate multiple VTUs at once.

The changes included in this PR should correct these issues. The idea is for the CI to stay green and for the reviewer(s) to generate a parallel model using this branch and assert that results can be properly loaded into ParaView and that pvtu2vtu works.

To review the changes, please have a look at the second commit attached; the first one only deals with formatting, no changes were applied.

### DIFF
--- a/femtools/VTK_interfaces.F90
+++ b/femtools/VTK_interfaces.F90
@@ -94,11 +94,11 @@ module vtk_interfaces
      end subroutine vtk_get_sizes
   end interface
 
-  interface vtkwriteghostlevels
-     subroutine vtkwriteghostlevels(ghost_levels)
+  interface vtkwritecellghostarray
+     subroutine vtkwritecellghostarray(ghosts)
        implicit none
-       integer ghost_levels(*)
-     end subroutine vtkwriteghostlevels
+       integer ghosts(*)
+     end subroutine vtkwritecellghostarray
   end interface
 
 contains
@@ -253,7 +253,7 @@ contains
     type(vector_field) :: v_model(3)
     type(tensor_field) :: t_model
     logical :: dgify_fields ! should we DG-ify the fields -- make them discontinous?
-    integer, allocatable, dimension(:)::ghost_levels
+    integer, allocatable, dimension(:) :: ghosts
     real, allocatable, dimension(:,:) :: tempval
     integer :: lstat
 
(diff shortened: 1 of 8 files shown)


gnikit (commented):
LGTM I would consider adding at a later point black and isort in the dev dependencies of setup.cfg /pyproject.toml along with some other things like precommit. I think there is a PR switching to setuptools #351 

I will however leave the approval of this PR to someone else.

angus-g (changes requested):
Interface changes look fine, but let's leave out the whitespace and formatting changes for now. As @gnikit suggested, the Black formatting could come into a separate CI change. I think the Fortran is sufficiently crufty that it's not worth doing a wholesale formatting change.

Patol75:
@angus-g This one is actually a massive pain to rebase, mostly because of the Python formatting changes. Any chance to proceed as-is?

angus-g (approved):
I guess so. It would be nice to get the large-scale formatting done then so we don't keep running into this.

Patol75:
Thanks, @angus-g.

I have just pushed a pre-commit branch. For now, all I have done is patch files using information detailed [here](https://pre-commit.com/). CI should follow based on [this](https://pre-commit.ci/). Actions run will be [here](https://github.com/FluidityProject/fluidity/actions/runs/2620971692).

angus-g:
Hmm, I rebased this onto main just to get the PR in, but it seems like the [new test failure](https://github.com/FluidityProject/fluidity/runs/7208629853?check_suite_focus=true#step:2:247) is at least somewhat real?

Patol75:
This test passed in the previous Actions run. Additionally, I remember seeing similar errors when I went through all the tests for the CMake build. Unfortunately, I do not remember what I did in that case if anything. I cannot see any major changes for that particular test in the CMake branch, and I do not think that the changes in the present branch affect this test at all. I am thinking it could well turn green if triggered again.

Patol75:
Thanks @angus-g, all green :)

Issue #115, opened 2016-05-03
CMake build for fluidity


jrper:
In a previous issue ( #113 ) @ggorman suggested migrating fluidity to use CMake as its build system would make some of our support issues (particularly VTK) easier. Over the long weekend I've been experimenting with getting a working build together (see the branch https://github.com/FluidityProject/fluidity/tree/cmake_build). It's not finished yet, and brought up some other issues I'll be questioning shortly, but to summarise my findings so far:

1. VTK is indeed really easy, and PETSc isn't to nasty either
2. The language syntax makes a bit little bit more sense as a new user than autoconf, and feels a bit more like a programming exercise
3. The cross-compiler and cross-platform support is good,

however, as @stephankramer said, debugging is annoying and it's unclear which of multiple routes is the best way to pass a given flag to a given compiler.

I'm assuming this will be a won't fix, but I thought I'd open this to further discussion, particularly in the case that there's interest in having an experimental or unsupported CMake build scheme available for people on weird systems.

stephankramer:
Had a quick look at that branch, and it looks great! Would be very interested to have a more detailed, offline discussion about how far you got.

tmbgreaves:
Thanks from me also - currently completely tied up with learning the intricacies of Jenkins plus docker but very interested to discuss / look at this more when time permits.

Patol75:
I am terribly late to the party here, but is there any chance for this Fluidity CMake build to get finalised? The reason why I am asking is that CMake has deprecated the command-line argument `--find-package` (e.g. [here](https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-the-find-package-tool) and [here](https://gitlab.kitware.com/cmake/cmake/-/issues/20849)) which is used in Fluidity's `configure` to retrieve some information about VTK. In particular, `configure` fails with the newer VTK 9:

```
configure: Checking for VTK automagic.
CMake Error at /apps/cmake/3.18.2/share/cmake-3.18/Modules/FindThreads.cmake:62 (message):
  FindThreads only works if either C or CXX language is enabled
Call Stack (most recent call first):
  /g/data/xd2/modules/vtk/9.1.0/lib64/cmake/vtk-9.1/VTK-vtk-module-find-packages.cmake:115 (find_package)
  /g/data/xd2/modules/vtk/9.1.0/lib64/cmake/vtk-9.1/vtk-config.cmake:150 (include)
[...]

PR #356, opened 2022-06-18
ENH: Discuss inclusion of a CMake build

### LINKED ISSUE #113: VTK library name changes

Patol75:
I am opening this PR to discuss a possible addition of a CMake build to Fluidity as an alternative to Autotools.
The main motivation is the deprecation by CMake of the command-line argument `--find-package` (more details [here](https://github.com/FluidityProject/fluidity/issues/115#issuecomment-965840649)) that was used in #125 to fix #113. This deprecation complicates building against recent VTK libraries using Autotools. Overall, it might be valuable to support CMake to avoid such issues, even though it is not impossible it brings issues of its own.

I will write the content of the PR later using the tree of file changes GitHub provides, summarising the changes I have made building on @jrper original build described in #115.

By no means the present PR should be merged as-is.


Patol75:
Given the massive changes included in the present branch, I have pushed another branch, called [cmake_build_2022_minimal](https://github.com/FluidityProject/fluidity/tree/cmake_build_2022_minimal), which only contains a minimal changeset to allow for a CMake build and unit tests to successfully complete. Changes can be found [here](https://github.com/FluidityProject/fluidity/commit/a039f5d365f1e2318c321b381857fe878ce3dbde), and the associated successful Actions run [here](https://github.com/FluidityProject/fluidity/actions/runs/2526262259).

To summarise the changes, there is now a `CMakeLists.txt` file within most of Fluidity's sub-directories that contain source files. Additionally, Python packages are now installed in the user site-packages directory and can be accessed within the Python interpreter without having to set the `PYTHONPATH` environment variable.

Steps to build Fluidity from a fresh Jammy 22.04 installation are detailed in `docker/Dockerfile.jammy`. Packages are downloaded from the official Ubuntu package repositories using the latest versions available (e.g. VTK 9.1), except for GMSH and PETSc. The former is downloaded from the official website as a pre-built binary, while the latter is first cloned from the official release branch hosted on GitLab and then built.

Configure- and build-time instructions are mostly found in the `CMakeLists.txt` file at the root of Fluidity's directory. Programs/packages marked as required include HDF5, MPI, NetCDF, PETSc, Python 3, sed, TCMalloc, UDUNITS and VTK. Additionally, a few libraries are fetched over the network: libspatialindex is downloaded from the official GitHub repository and seamlessly integrated into Fluidity's build as it also uses CMake, while Judy and Zoltan are downloaded from their respective SourceForge and GitHub repositories and built according to provided instructions. The only environment variables required to be set by the user are `PETSC_ARCH` and `PETSC_DIR`.

Questions, comments, feedback and suggestions are all very much welcomed.

PR #357, opened 2022-06-20
Update autoconf macros for Python, BLAS and LAPACK

### LINKED ISSUE #355: Incorrect check for Python3 version
In the autoconf-generated section of configure there's a problematic version comparison:

https://github.com/FluidityProject/fluidity/blob/86ccd0171513197a174653dfb1e80c086058b3d2/configure#L6141

This is fine until 3.10, at which point it fails, with "3.10" >= "3.6" returning false.

The check should probably be using parse_version from pkg_resources, ie:

parse_version("3.10") >= parse_version('3.6')

I can't immediately think of a quick and sane way to deal with this as it's not in our code, but from autoconf. 

Thoughts?

angus-g:
Using autoconf-archive 2022.02.11, and regenerated configure script using autoconf 2.71. `AX_BLAS` and `AX_LAPACK` are modified to not mangle the routine names, since they're not being cross-linked. Hopefully it's not too much of a stretch to make these things a bit more compatible with more modern toolchains?

Closes #355.


stephankramer (approved):
Looks great, thanks!

PR #360, opened 2022-07-11
ENH: Automatically fix a number of formatting issues using pre-commit

Patol75:
For this PR, I have used the [pre-commit](https://pre-commit.com/) framework to automatically check and correct formatting issues within the `main` branch of Fluidity. Hooks used are listed in `.pre-commit-config.yaml` and detailed [here](https://pre-commit.com/hooks.html). I have only encountered two issues. The first one was with `Makefile.dependencies` files, which show a conflicting end-of-file behaviour when they are generated through `make makefiles`. The second one is with regard to a binary VTU file for which I have not investigated further. None of these two issues seems significant to me.

`pre-commit` also provides a [CI solution](https://pre-commit.ci), but unfortunately, it is not available for free for GitHub organisations. In any case, I believe the automatic changes included here will be helpful with future development of Fluidity as formatting should not be an issue anymore.


stephankramer:
This is great. Tried it out for myself and seems to work nicely and non-intrusively.

.github/workflows/ubuntu.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - pre-commit
```
angus-g, line 7: Probably drop this before merge?

angus-g:
Looks like a good idea. I think the CI Action should still be available (it's for "open source repositories"), it would just need a FluidityProject admin to install it, I assume.

Patol75:
Thank you @angus-g and @stephankramer.

@stephankramer Any chance to try the suggestion of @angus-g? You just need to log in using Fluidity's GitHub account [here](https://pre-commit.ci/) and see if it lets you install pre-commit for the repository for free. I have done it with my repositories, and it was pretty straightforward.

angus-g:
Looks like the change to vtutools.py is breaking CI:
```diff
diff --git a/python/fluidity/diagnostics/vtutools.py b/python/fluidity/diagnostics/vtutools.py
index ad7c4ba0a..f89798a05 100644
--- a/python/fluidity/diagnostics/vtutools.py
+++ b/python/fluidity/diagnostics/vtutools.py
@@ -28,8 +26,9 @@ import fluidity.diagnostics.elements as elements
 import fluidity.diagnostics.optimise as optimise
 import fluidity.diagnostics.simplices as simplices
 import fluidity.diagnostics.utils as utils
+import numpy as np
 import vtk
-from vtktools import *
+from vtktools import VtuMatchLocations, vtu
 
 
 def VtkSupport():
```

Lots of tests import vtutools as vtktools, which is a bit confusing!

Patol75:
That is a good break, it allows us to see which tests import `vtutools` as `vtktools`, making use of the now-deleted `from vtktools import *` within vtutools. I am pretty sure I have already fixed these issues, but very likely in another branch. I will wait until CI finishes, commit the additional changes to the test, and we should be good to go.

Patol75:
CI was all green before the last commit, so all good there.

Thanks, @gnikit for the comments. I had indeed first tried `isort`, but I was unaware of the compatibility arguments with `black`, hence why I had tried another hook. That should be sorted now. For `flake8`, well, I had already tried it too, but given it only highlights problems and does not fix them, I had left it out. Oh well, now it is in, and it does not seem to create much trouble. Also, thanks for the sanity check!

Given the latest commit here, I see that `pre-commit` is now part of the CI. Thanks, @stephankramer. However, this brings back the issue with `tests/data/ref_vtk_hexahedra.vtu`: expect a failure from unit tests. Should we attempt to fix this file, or simply exclude it from the `mixed-line-ending` hook? The latter would look like that:
```
    -   id: mixed-line-ending
        exclude: ref_vtk_hexahedra.vtu
```

gnikit:
I think I added that file a while back when unittesting the node ordering of the VTK routines. Do we know why the test fails? I am looking at a diff and the two files and they look the same to me.

stephankramer:
@gnikit : out of curiosity, any particular reason you prefer isort over reorder_python_imports?

gnikit:
> @gnikit : out of curiosity, any particular reason you prefer isort over reorder_python_imports?

Substantially wider adoption from projects so less likely to be abandoned (isort 460M vs reorder_python_imports 831k). `reorder_python_imports` is owned by an individual whereas `isort` is owned py PyCQA

Patol75:
CI should now return all green on the building and testing sides, but [pre-commit is unhappy](https://results.pre-commit.ci/run/github/20023221/1657760140.oB_1HY73TeOuxSRaMB-S4g) because the compile symlink in libspatialindex is broken. I updated it in the original commit because it was broken in both Focal and Jammy, but now it looks like it is broken in the CI. Am I missing something, or should I just exclude that file from the broken-symlink hook?

stephankramer:
Yeah, I'm not sure I ever intended to commit that symlink. Looks like an odd thing to do ¯\_(ツ)_/¯
There's still the flake8 stuff of course, which I think is a good addition but we may need some more exceptions (after fixing the obvious)

Patol75:
Thanks, @stephankramer. I have merged the warnings branch here and started to address the flake8 complaints. There are quite a lot of them, so that is going to take some time. Any help appreciated.

Patol75:
PR should now be good to go.

Patol75:
Thanks for the changes @stephankramer.

PR #361, opened 2022-07-15
Jammy updates

stephankramer:


Patol75:
@stephankramer Thanks for putting this together, nice to see that things run smoothly on Jammy. However, a small heads-up that Impish reached end-of-life, so it should now be removed from the CI.

gnikit (dismissed):
Can we merge this and drop Impish support in another PR? There is some stuff happening with pre-commit and Bionic so I think @stephankramer you might have to rerun the CI.

PR #367, opened 2022-11-30
Update versions of actions to avoid deprecation warnings

Patol75:
GitHub has made some changes to Actions (links below), and as a result, versions of actions used in our workflows should be updated.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -32,17 +32,17 @@ jobs:
     steps:
 
       - name: Check Out Repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Login to Docker Hub
-        uses: docker/login-action@v1
+        uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_TOKEN }}
 
       - name: Build and push
         id: docker_build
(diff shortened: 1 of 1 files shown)


Patol75:
CI is all green, please feel free to approve and merge.

stephankramer (approved):
Great thanks

gnikit (approved):
This looks good. I will add some automation in a future PR to not have to worry again.

gnikit:
@stephankramer any idea what's happening with the Bionic checks? As far as I can see, `Unit Bionic` and `Short Bionic` seem to be missing from `ubuntu.yml`:
`testing > strategy > matrix > name` and `include`

stephankramer:
Yes, that's weird. I've removed these from main - and if you look at https://github.com/FluidityProject/fluidity/actions/runs/3581331774 it's seems all happy and green (without these two as intended) - it's only on this PR that it seems to insist on these still being run? I'll just override and merge it regardless - hopefully we won't see this come back on other PRs

stephankramer:
Yes, thanks, you were right: it specifies explicitly which statuses are required for a merge to be allowed. Removed the Bionic medium/unit ones now and added the Jammy ones

PR #383, opened 2023-11-23
Fixed bug in linked list where the lastnode wasn't being updated when…

stevendargaville:
… the last value was popped off the end of the list. Causes a segfault the next time you try and access the lastnode after a pop_last.

### DIFF
--- a/femtools/Linked_Lists.F90
+++ b/femtools/Linked_Lists.F90
@@ -353,6 +353,7 @@ contains
     deallocate(node)
     prev_node%next => null()
     list%length = list%length - 1
+    list%lastnode => prev_node
   end function ipop_last
 
   function ifetch(list, j)


stephankramer:
I'm a little confused: if you don't update `lastnode` then it presumably still points to `node` (the last current node it has found by looping through the list), which has now been deallocated?

The issue I see with the current code is that if the length of the current list is one, it never goes into the loop, so `prev_node` is never initialised. There should probably be a special case where it leaves a valid 0-length list behind in that case.

stevendargaville:
> I'm a little confused: if you don't update `lastnode` then it presumably still points to `node` (the last current node it has found by looping through the list), which has now been deallocated?

Yes, so I think that if you call pop_last and then immediately call insert for a non-zero length list, line 264 (node => list%lastnode) in insert will segfault? 

> The issue I see with the current code is that if the length of the current list is one, it never goes into the loop, so prev_node is never initialised. There should probably be a special case where it leaves a valid 0-length list behind in that case.

Yes you're right, there should be a guard on the assignment of lastnode with a list of size one, I'll add that now.

PR #393, opened 2026-03-24
Try CI with current gmsh version

stephankramer:


stephankramer:
So the assess issue has been resolved. Assess now needs scipy>=1.15 but wasn't advertising this and Noble is still on scipy 1.11. I've fixed this requirement in assess, and so now `pip install assess` pulls in a newer version of scipy automatically.

@Patol75 : would you mind taking a look at the failing particle_rayleigh_taylor_mu10 example as you and Chris were involved in this? It appears to be checking some values against what looks like .stat files (bench-25.dat and bench-48.dat) from a previous run? In any case, the number it checks appears to be mesh-dependent. I'm inclined to just delete these two (of the four) acceptance tests - but will let you decide...

Then particle_rayleigh_taylor_checkpoint is failing, which as mentioned elsewhere seems to be a bit flaky (already on gmsh 2) so I'll just loosen its tolerance a bit.

Then I still need to rebuild the Jammy base container to also run with gmsh4 - or I could just drop it altogether....

stephankramer:
Looks like switching to gmsh 4 in Jammy, or just the rebuild of the Jammy base container triggered all sorts of new issues (to do with numpy C API incompatibilities?). So I'm just going to drop Jammy support altogether as that was planned anyway.

PR #395, opened 2026-03-27
Fixes #394: Update Python setup

benfulton:
Now using setuptools instead of distutils. The CI does not build python so testing must be done manually by running the configure as 

```
HAVE_PYTHON=yes ./configure --enable-2d-adaptivity
```


Patol75 (commented):
These changes look reasonable to me. I see a few CI failures, but they are likely not related to this PR. Are the spherical benchmark longtest failures linked to changes in `assess`? `particle_rayleigh_taylor_checkpoint` seems to fail intermittently, so nothing too concerning there.

EDIT: Yes, spherical failures are due to `assess` (https://github.com/FluidityProject/fluidity/actions/runs/23657117725).

PR #396, opened 2026-06-09
Update to a more modern BLAS detector

### LINKED ISSUE #294: Configure not picking up BLAS and LAPACK from PETSc
This is more of an open-ended question, but I think it might be worth pointing it out.
I have been playing with PETSc and Intel's new OneAPI toolkit to try and speed some of my local calculations. In doing so I realised that when using gfortran Fluidity picks up BLAS & LAPACK only from their standard library locations i.e. when installed through APT, which does not make much sense since we can also install them through PETSc and that should also be picked up.

Moreover, attempting to specify the location of BLAS & LAPACK with `--with-blas=` and `--with-lapack=` will fail because these options are not transferred over when `./configure` is called for `libadaptivity`, which in turn will attempt to pick up BLAS & LAPACK from the system location, causing it either to fail or link against a different set of libs (possibly true for Judy as well).
This can be circumvented by passing to the `./configure` `LIBS="-L$PETSC_DIR/$PETSC_ARCH/lib":${LIBS}`, but then the real question is why have `--with-blas`, `--with-lapack` if they do not work as expected and in some cases can lead to Fluidity using multiple versions of BLAS & LAPACK?

benfulton:
The BLAS autoconf detector macros are out of date. This causes issues with libraries like FlexiBlas which is provided with EasyBuild - it will be simpler to find libraries with these changes. The detection will have to run in C mode rather than Fortran mode but that shouldn't cause any issues. I think this fixes #294.

Updated to the latest versions from [autoconf-archive](https://github.com/autoconf-archive/autoconf-archive) and run autoconf to generate a new configure.


stephankramer (approved):
Looks good!

Just to double check:
* is it indeed the case that in libadaptivity we only call blas in libadaptivity/include/cxxblas.h via `extern "C"`, i.e. we don't need to test for fortran or CXX
* the ax_blas/lapack that are used by fluidity's main configure is already up-to-date (enough)?

benfulton:
Yes, it's only in the C++ code with `extern "C"` that BLAS is used, and the main configure is already using these macros.

PR #398, opened 2026-06-09
Small changes to make configuration more pleasant

benfulton:
- Update some CI version tools

- Remove FLIBS and LAPACK_LIBS from the climatology makefile. (They are redundant and can cause problems with overlong command lines)

- Change femtools to support NumPY 2.0

- Update Python installer to use `python -m pip install` rather than `python setup.py`

- Remove clean_python from the clean command in the makefile as it no longer exists


femtools/embed_python.c
```diff
@@ -43,6 +43,17 @@ USA
 #ifdef HAVE_NUMPY
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
 #include "numpy/arrayobject.h"
+
+static int ensure_numpy_api(void) {
+  if (PyArray_API != NULL) {
+    return 0;
+  }
```
stephankramer, line 51: Could you add a comment what this does? Is it checking whether numpy has been "imported" already?

stephankramer (dismissed):
All good - maybe add a comment

stephankramer (approved):

PR #402, opened 2026-07-01
Install NumPy < 2.0 along with Assess

benfulton:
Apparently the Assess install in the CI Dockerfile has started to bring in NumPy 2.0. This should work in the general case, but many tests rely on earlier versions of NumPy and will fail under NumPy 2.0. Additionally, since we compile the application under Numpy 1.0, updating it to 2.0 at the end will cause problems. The solution is to install NumPy<2.0 alongside Assess at the end of the dockerfiles.

### DIFF
--- a/docker/actions/Dockerfile.actions.noble
+++ b/docker/actions/Dockerfile.actions.noble
@@ -35,4 +35,4 @@ RUN make manual
 RUN sudo make install
 
 # Python module 'assess' is required for some longtests
-RUN python3 -m pip install --break-system-packages assess
+RUN python3 -m pip install --break-system-packages "numpy<2" assess
--- a/docker/actions/Dockerfile.actions.noble-omp
+++ b/docker/actions/Dockerfile.actions.noble-omp
@@ -34,4 +34,4 @@ RUN make fltools
 RUN make manual
 
 # Python module 'assess' is required for some longtests
-RUN python3 -m pip install --break-system-packages assess
+RUN python3 -m pip install --break-system-packages "numpy<2" assess


stephankramer (approved):
I see, makes sense. Any idea why some tests fail with numpy 2? We should probably fix that as it's more and more likely to be what users have

benfulton:
NumPy 2 introduced a lot of breaking changes, e.g. size-1 arrays like np.array([3.14]) are no longer implicitly convertible to Python scalars. A lot of the tests rely on this - particle_dt_test is an example:

```
import numpy as np
import h5py
from fluidity_tools import stat_parser

f = h5py.File('particle_attributes.particles.DtMeasure.h5part', 'r')
z = stat_parser("particle_attributes.stat")
dt_errors = np.zeros(10,float)

for i in range(10):
    d = f['/Step#{}'.format(len(f)-(10-i))]
    val = d['DtVal'][:]
    dt_errors[i] =  val- z['dt']['value'][i+1]
dt_error=abs(dt_errors).max()
print ("max error",dt_error)
Variable computation raised an exception
--------------------------------------------------------------------------------
  1  import numpy as np
  2  import h5py
  3  from fluidity_tools import stat_parser
  4  
  5  f = h5py.File('particle_attributes.particles.DtMeasure.h5part', 'r')
  6  z = stat_parser("particle_attributes.stat")
  7  dt_errors = np.zeros(10,float)
  8  
  9  for i in range(10):
 10      d = f['/Step#{}'.format(len(f)-(10-i))]
 11      val = d['DtVal'][:]
 12      dt_errors[i] =  val- z['dt']['value'][i+1]
 13  dt_error=abs(dt_errors).max()
 14  print ("max error",dt_error)
--------------------------------------------------------------------------------
TypeError: only 0-dimensional arrays can be converted to Python scalars

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/bfulton/GIT/fluidity/python/fluidity/regressiontest.py", line 363, in run_python
    exec(self.code, varsdict)
    ~~~~^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 12, in <module>
ValueError: setting an array element with a sequence.
--------------------------------------------------------------------------------
particle_dt_test: failure.

particle_dt_test.xml: F
```

PR #408, opened 2026-07-13
Support HDF5 1.12 directly in the source

benfulton:
The current configure has code to patch the h5hut directory in the presence of HDF5 version 1.12. This was done, presumably, in order to avoid touching the h5hut source which was written by a 3rd party. Unfortunately the patch no longer works with HDF5 1.14, and h5hut appears to be no longer maintained. This PR removes the patching code from configure and modifies the h5hut code directly to work with versions of HD5 up to 1.14.6.

### DIFF
--- a/configure
+++ b/configure
@@ -14983,51 +14983,9 @@ if test "$with_hdf5" = "no"; then
 fi
 LDFLAGS="$LDFLAGS $HDF5_LDFLAGS"
 
-# Some changes are necesary in H5hut source to build with HDF5 >= 1.12
-
-
-
-  # Used to indicate true or false condition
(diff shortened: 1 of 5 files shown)


stephankramer (approved):
Looks good. Unfortunate that h5hut seems abandoned indeed. This came up recently with @angus-g - looking for a scalabel checkpointing mechanism for (moving) particles in firedrake.

angus-g:
> Unfortunate that h5hut seems abandoned indeed

Ah that's annoying! I think Will said he's using h5hut or h5part for his stuff, so I wonder if it's through the same library or a different interface...

PR #409, opened 2026-07-29
Fixes to make test runs more generic

benfulton:
* Some mpiexec implementations resolve duplicate -n flags unpredictably. Modify the testharness to not add a -n flag if one already exists; and modify a couple of tests to pass the flag explicitly.

* Relax the gauss_point_buoyancy test slightly as it gives slightly different results on an HPC machine

* Set up serialtest so it does not require spudtools to be installed before running

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -441,8 +441,8 @@ distclean: clean
 
 test: serialtest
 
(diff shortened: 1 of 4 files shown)


stephankramer (approved):
Marvellous, looks good

